PR #1, opened 2015-06-14
Use pods for external depedencies

Deraen:
Hello,

This is a quite large change which removes all set-env calls from namespaces, moves tasks to one namespace and loads the external dependencies using [pods](https://github.com/boot-clj/boot/wiki/Pods) for the tasks which require those.


podviaznikov:
Thanks for the PR! I'm going to read about pods more.
I wanted to have all task/plugins in the separate file because potentially they can be removed from perun into separate repos. I was thinking that ideally perun should be minimal with everything implemented as plugin. I didn't explore that idea much, but I think you tried to bundle most of the tasks into the single file. Let me know what you think. I don't have final idea how everything should be separated it, so I open to exploration and experiments.

Deraen:
I don't see much value in splitting the tasks to separate repos unless they grow much more complex. Because of pods Perun doesn't have any direct dependencies. Clj-time and such are only loaded if user uses a task which needs one of those dependencies.

podviaznikov:
Basically my main idea why I'd like to see independent tasks is that people can create tasks on their own and chain it together with other tasks to build custom static site generator. And basically I don't want people depend on me to merge their task in perun.
http://www.metalsmith.io/ was one of my inspirations. They have a lot of 3rd party reusable plugins.
I'm ok with having several core tasks, but I think it would be great if perun can be extensible.

Deraen:
This change doesn't make Perun any less extensible :)

Though some thought could be given on providing some helper functions (around find-data-file) to make creating new tasks easier.

Deraen:
Note: I'm not perfectly happy e.g. with find-data-file and read-file-defs, but I'm thinking changes in #2 should make both of those pretty much unnecessary.

As metadata is presentable in EDN it should be easy to pass to the pods which would remove loads of file path passing.

podviaznikov:
Agree that it doesn't changes extensibility of perun. I definitely like what you did with pods. That is super useful. I'm still not sure how I feel about putting all of the tasks into the one file. 
Were there some practical reasons of doing so? I see that now we have global defaults and task defaults. I like that.
I think intuitively I like smaller files. If you can explain what it gives us to have all tasks in one file/namespace - that would help.

Deraen:
And the reason why I moved the tasks from their separate namespaces is that implementation namespaces (e.g. io.perun.rss) use dependencies which are not available root classloader (outside of pods). And inside of pods e.g. `boot.core` is not available.

It would still be possible to separate each task to own namespace, but for tasks which use external dependencies there should be another impl namespace which is only loaded inside pod.

podviaznikov:
Ok, thanks for explanation and PR. Merging!

podviaznikov:
agree about your note on `find-data-file` and `read-file-defs`. Going to work on removing `meta.edn`

PR #6, opened 2015-06-14
look for .markdown extension in task

martinklepsch:
Not 100% convinced about looking for multiple file extensions but given that the maintenance overhead of this is zero, it's probably fine.

Something else I found on the topic: http://daringfireball.net/linked/2014/01/08/markdown-extension


podviaznikov:
Thanks, this makes sense

Issue #2, opened 2015-06-14
Faster recompilation


Deraen:
Currently when using this together with watch task with 20 posts recompilation takes maybe 1.5 seconds. This could probably be made much faster by only parsing the changed md files.

Idea:
- Instead of writing metadata to a file, assoc it to fileset metadata
- Store metadata as a map of md file -> metadata
- Markdown task would check for changed files and update those in the metadata

podviaznikov:
It's and interesting idea to store everything in fileset  metadata.
Are you proposing to get read of `meta.edn` completely?

Deraen:
To get rid of meta.edn. Yes.

But the more important change would be to store metadata in map and parse only changed files. Using fileset metadata is not required for this.

podviaznikov:
And how would you pass that map between boot tasks? You want to have global map accessible from all the tasks?

Deraen:
If I remember correctly, this should work:

``` clj
(boot/with-pre-wrap fileset
  (println (:metadata (meta fileset)))
  (with-meta fileset {:metadata {"/foobar.md" {}}}))
```

podviaznikov:
I think I like that. Never used `meta` before but it's seems like a really good use-case to try.
Let me play with this

Deraen:
Check https://github.com/Deraen/boot-less/blob/master/src/deraen/boot_less.clj#L29-L32 for example how to find which files changed runs.

podviaznikov:
I have PR (https://github.com/hashobject/perun/pull/11) to store all metadata as map md->file metadata (as was suggested).

podviaznikov:
I was thinking about this and now faster recompilations makes more sense. I saw how you are using perun with http-server and livereload and it makes sense that you want faster reloads.
My initial workflow was different and I didn't care about speed before. But I think I like your workflow more. And I like that boot just allows us to chain those tasks together.

podviaznikov:
Fixed in https://github.com/hashobject/perun/pull/15

PR #7, opened 2015-06-14
Remove meta.edn

podviaznikov:
All metadata stored now as clojure metadata associated with boot fileset.

### DIFF
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -114,11 +114,11 @@
         tmp (boot/tmp-dir!)
         options (merge +sitemap-defaults+ *opts*)]
     (boot/with-pre-wrap fileset
-      (let [datafile (find-data-file fileset (:datafile options))]
+      (let [files-metadata (:metadata (meta fileset))]
         (pod/with-call-in @pod
           (io.perun.sitemap/generate-sitemap
             ~(.getPath tmp)
-            ~(.getPath (boot/tmp-file datafile))
+            ~files-metadata
             ~options))
         (commit fileset tmp)))))
 
@@ -142,11 +142,11 @@
         tmp (boot/tmp-dir!)
         options (merge +rss-defaults+ *opts*)]
     (boot/with-pre-wrap fileset
-      (let [datafile (find-data-file fileset (:datafile options))]
+      (let [files-metadata (:metadata (meta fileset))]
         (pod/with-call-in @pod
           (io.perun.rss/generate-rss
             ~(.getPath tmp)
-            ~(.getPath (boot/tmp-file datafile))
+            ~files-metadata
             ~options))
         (commit fileset tmp)))))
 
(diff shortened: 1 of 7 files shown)


martinklepsch:
cc @Deraen

podviaznikov:
@Deraen I removed `meta.edn`. It's one of the changes you suggested in #2. Structure of metadata is the same for now. I'll probably do another PR for that.

Let me know if you have comments about this PR.

podviaznikov:
Reminder to myself: still need to update reamde.

src/io/perun.clj
```diff
@@ -36,50 +36,48 @@
...
-        (commit fileset tmp)))))
+      (let [markdown-files (->> fileset boot/user-files (boot/by-ext ["md" "markdown"]) (map #(.getPath (boot/tmp-file %))))
+            parsed-metadata (pod/with-call-in @pod
+                              (io.perun.markdown/parse-markdown
+                                ~(.getPath tmp)
+                                ~options
+                                ~markdown-files))
+            fs-with-meta (with-meta fileset {:metadata parsed-metadata})]
+        (u/info "Parsed markdown files\n")
+        (commit fs-with-meta tmp)))))
```
Deraen: Markdown task is not writing to any files? tmp-dir and commit can be removed.

Deraen:
Other than my comment about markdown task, looks good!

src/io/perun.clj
```diff
@@ -36,50 +36,48 @@
 
 (deftask markdown
   "Parse markdown files"
-  [d datafile        DATAFILE        str "Target datafile with all parsed meta information"
-   f create-filename CREATE_FILENAME str "Function that creates final target filename of the file"]
+  [f create-filename CREATE_FILENAME str "Function that creates final target filename of the file"]
```
martinklepsch, line 39: not really related to this PR but `create-filename` does not seem to be used anywhere?

src/io/perun.clj
martinklepsch: but it adds metadata — does that not require `commit!`?

src/io/perun.clj
Deraen, line 39: It's used in markdown/process-file

podviaznikov, line 39: It's used here: https://github.com/hashobject/perun/blob/remove-meta-edn/src/io/perun/markdown.clj#L55. Passed as option.

src/io/perun.clj
Deraen: Nope. Commit is to write changes to disk. Clojure metadata is just associated with Java object.

podviaznikov: Yeah, just double checked. `commit!` is not required. Thanks for the catch

podviaznikov:
Thanks for the review @Deraen and @martinklepsch

Issue #8, opened 2015-06-14
Parse markdown metadata specified in Jekyll style


podviaznikov:
See http://jekyllrb.com/docs/frontmatter/

podviaznikov:
Would be easier when if this task is done: https://github.com/yogthos/markdown-clj/issues/46

podviaznikov:
@martinklepsch this is now in master. So you should switch to YAML metadata now.

martinklepsch:
Great! Nice work!

PR #10, opened 2015-06-16
change format of markdown metadata

podviaznikov:
Implementation for the https://github.com/hashobject/perun/issues/8.

Now markdown metadata should be specified in the Jekyll style.
It's good because GitHub hash special support for such style.

### DIFF
--- a/build.boot
+++ b/build.boot
@@ -6,7 +6,6 @@
                   [jeluard/boot-notify "0.1.2" :scope "test"]
                   [clj-time "0.9.0" :scope "test"]
                   [markdown-clj "0.9.40" :scope "test"]
-                  [endophile "0.1.2" :scope "test"]
                   [time-to-read "0.1.0" :scope "test"]
                   [sitemap "0.2.4" :scope "test"]
                   [clj-rss "0.1.9" :scope "test"]])
@@ -14,7 +13,7 @@
 (require '[adzerk.bootlaces :refer :all])
 
 
-(def +version+ "0.1.0-SNAPSHOT")
+(def +version+ "0.1.1-SNAPSHOT")
 (bootlaces! +version+)
 
 (task-options!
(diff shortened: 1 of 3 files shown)


Deraen:
You should also remove endophile dependency from io.perun/markdown-deps.

Does the Jekyll specify if metadata has to be between LINES containing only "---"? Current solution would break if metadata contained "---". Regex could be modified to match only lines not containing anything else: `#"^---$"`.

I might use `line-seq` to read the string, skip elements before first "---" line and then take lines before the second "---". Using `line-seq` would also have a small(?) performance benefit of not requiring to read the whole file ~~or needing to store complete file in a string.~~ the whole file is anyway read for content so probably this wouldn't be much better.

Deraen:
If would be good practice to add `:require` for `clojure.string` (with alias maybe).

podviaznikov:
endophile is removed now, updated use of clojure.string. I like suggestion about `line-seq`.

podviaznikov:
Agree about problem when metadata contains `---`

Deraen:
After thinking about this a bit more, I'm not so sure if `lazy-seq` is better. As the whole content needs to be read to a string anyway the slurp is quite good fit.

podviaznikov:
Head the same conclusion about `line-seq` after checking that we read file anyway later.

podviaznikov:
I changed to read file only once. Used YAML parser library and improved how we search for YAML inside the markdown.

Deraen:
`string/split` will throw a NPE if the string is nil, which happens if first split doesn't return second item.

`some->` should fix that.

Deraen:
And remove-metadata also should support case where the file doesn't have metadata. I'm not quite sure what would be best solution there.

podviaznikov:
Thanks for the comments on potential issues. Going to work on it now.

podviaznikov:
I'm using `some->` now. Thanks for the tip. I wasn't aware of this macro.
Also right now by default each file has empty metadata. So removing would still work.

Deraen:
:+1:

PR #11, opened 2015-06-16
Change metadata format

podviaznikov:
New format was proposed in #2.

Now instead of list of meta-data entries for all files we have a map like this:

```
{"/post1.md" {:name "Post 1"} "/post2.md" {:name "Post 2"}}
```

### DIFF
--- a/build.boot
+++ b/build.boot
@@ -9,13 +9,12 @@
                   [endophile "0.1.2" :scope "test"]
                   [time-to-read "0.1.0" :scope "test"]
                   [sitemap "0.2.4" :scope "test"]
-                  [clj-rss "0.1.9" :scope "test"]
-                  [metosin/potpuri "0.2.2" :scope "test"]])
+                  [clj-rss "0.1.9" :scope "test"]])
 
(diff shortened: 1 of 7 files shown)


Deraen:
`map-vals` (nearly same the as update-map) and `filter-vals` should be quite useful: https://github.com/metosin/potpuri/blob/master/src/potpuri/core.cljx#L203-L240 (You could copy the code)

podviaznikov:
Added `potpuri` as a dependency.

Deraen:
That doesn't work. You are trying to use potpuri in `io.perun` ns which is run in `build.boot` context where the library is not available. That is why I suggested you to copy the code.

podviaznikov:
and there is no other way to make it available inside the `build.boot` context?

Deraen:
Nope. The code run the `build.boot` context shouldn't depend on other than stuff provided by Boot (and Clojure).

podviaznikov:
Interesting, I didn't know that.

podviaznikov:
Going to merge. I increased the version so it's shouldn't break anything. Let me know if you'll have more comments.

PR #13, opened 2015-06-17
replace markdown-clj with endophile

martinklepsch:
This worked as a drop in replacement for me, properly processing and outputting many things `markdown-clj` did not parse as expected.


podviaznikov:
Thanks! Tested it and it works for me

PR #15, opened 2015-06-18
parse markdown only for changed files

podviaznikov:
Implementation for the last piece from #2


podviaznikov:
@Deraen check this out. It's much faster now. Small blogs are super fast now for me - 0.177ms. My big blog build time is down from 6s to 0.9s.

Deraen:
:+1:

PR #18, opened 2015-06-18
kebab-case for all metadata keys; closes #17

### LINKED ISSUE #17: look for metadata with dashes by default
Instead of looking for metadata like `date_published` be default it seems more idiomatic to look for `date-published`. As far as I understand both are valid YAML keys.

Happy to create a PR if wanted.

martinklepsch:


podviaznikov:
Thanks!

PR #19, opened 2015-06-18
remove +defaults+

martinklepsch:
Besides removing `+defaults+` I also removed some options related code around permalink generation which was and still is broken I believe.

Will probably look into that later. The URL scheme I personally want to achieve is:

```
/posts/title.html
```

I think there should be a sufficiently flexible mechanism to allow various schemes.

### DIFF
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -8,9 +8,6 @@
 (def ^:private global-deps
   '[[clj-time "0.9.0"]])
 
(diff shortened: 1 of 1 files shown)


podviaznikov:
Thanks!
My ideal URL schema is `/post-name/index.html`, this way it can be resolved in browser like `/post-name`. But permalink plugin should definitely support multiple schemes.

PR #21, opened 2015-06-18
Customizable permalinks

martinklepsch:
Permalinks should be easily customizable. The idea here is that the `permalink` task gets a function that determines the permalink based on the metadata of a file. We could create a slug based on the title of a file but that is complex and actually we already have a perfect slug at hand: the filename itself.

In most cases though the filename contains dates and other unwanted things. Thats where the `slug` task comes in: it adds a `:slug` key to the metadata based on the result of a `slug-fn` on the filename. `slug-fn` can be overwritten via task options of course.

Let me know what you think


src/io/perun.clj
```diff
@@ -74,18 +73,43 @@
...
+       (drop 3)
+       drop-last
+       (string/join "-")
+       string/lower-case))
+
+(deftask slug
+  "Adds :slug key to files metadata. Slug is derived from filename."
+  [s slug-fn SLUGFN sym "Function to build slug from filename"]
+  (boot/with-pre-wrap fileset
+    (let [slug-fn       (or slug-fn default-slug-fn)
```
Deraen, line 87: If slug-fn is a symbol it won't work as function?

martinklepsch, line 87: Oh right, that probably should be `code`.

podviaznikov:
I'm going to go through this PR in few hours. Looks good so far!

src/io/perun.clj
```diff
@@ -15,23 +16,23 @@
...
+
 (deftask markdown
-  "Parse markdown files"
-  [f create-filename CREATE_FILENAME str "Function that creates final target filename of the file"]
+  "Parse markdown files
+
+  This task will look for files ending with `md` or `markdown`
+  and add a `:content` key to their metadata containing the
+  HTML resulting from processing the markdown file's content"
+  []
```
podviaznikov, line 34: :+1: for removing `create-filename` from markdown task

src/io/perun.clj
```diff
@@ -74,18 +73,43 @@
...
+       (string/join "-")
+       string/lower-case))
+
+(deftask slug
+  "Adds :slug key to files metadata. Slug is derived from filename."
+  [s slug-fn SLUGFN code "Function to build slug from filename"]
+  (boot/with-pre-wrap fileset
+    (let [slug-fn       (or slug-fn default-slug-fn)
+          files         (:metadata (meta fileset))
+          updated-files (into {}
```
podviaznikov: You can use `https://github.com/hashobject/perun/blob/master/src/io/perun/core.clj#L36` function here. Will be a little bit shorter

src/io/perun.clj
```diff
@@ -138,11 +162,11 @@
         (commit fileset tmp)))))
 
 (def ^:private +render-defaults+
-  {:target "public"})
+  {:out-dir "public"})
```
podviaznikov, line 165: :+1: for renaming

src/io/perun.clj
```diff
@@ -151,22 +175,22 @@
         (doseq [file files]
           (let [render-fn (:renderer options)
                 html (render-fn file)
-                page-filepath (str (:target options) "/"
-                                   (or (:filepath file)
+                page-filepath (str (:out-dir options) "/"
+                                   (or (:permalink file)
```
podviaznikov, line 179: :+1: for making permalink prop

podviaznikov:
Looks much cleaner now. I like how each task does one job now.

PR #25, opened 2015-06-19
Change collection comparator default

martinklepsch:
Currently the default is `(fn [i1 i2] (compare i1 i2))`. With the assumption that in most cases this function will get two date strings like `20130512` or `2015-03-10` and people usually want to sort those most-recent-first I'd suggest changing the comparator function to `(fn [i1 i2] (compare i2 i1))`.


podviaznikov:
Agree with the new default. Thanks!

PR #29, opened 2015-06-20
Run render functions in new pods each run.

### LINKED ISSUE #3: Reload blog specific Clojure code
If I change e.g. post or collection render function I have to restart boot to see the changes. I can think of two alternatives to make this work with watch task so that the changes are picked up instantly.
1. A task witch uses clojure.tools.namespaces refresh
2. render and collection tasks are instead run inside a new pod each time

I would probably prefer the second option. It should be quite fast when used with pod pool.

Deraen:
Closes #3

This uses pod-pools to run render stuff in new pods each time, thus reloading the clojure code always. This requires that the rendered option is specified as fully qualified symbol resolving to a symbol, because we have to be able to load the namespace.

It is a bit slower than before. I'm thinking I'll add check to only use new pod only when either cljc or clj files have been changed.

### DIFF
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -179,6 +179,20 @@
 (def ^:private +render-defaults+
   {:out-dir "public"})
 
+(defn- wrap-pool [pool]
+  (let [prev (atom nil)]
+    (fn [fileset]
+      ; Do not refresh on the first run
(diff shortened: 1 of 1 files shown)


podviaznikov:
Looks goods. Now we finally have fully working live reload.

PR #40, opened 2015-06-25
Batteries for templates

podviaznikov:
Include some batteries for rendering using template engine. Instead of requiring rendering function we accept two properties: `:template-engine` and `:template-name`.

Starting with https://github.com/yogthos/Selmer.
Plan is to add few more: Jade,  Mustache.


podviaznikov:
@martinklepsch, @Deraen let me know what do you think about optionally adding support for the template engines.

Deraen:
I don't see myself using other than hiccup.

Instead of making `render` a monolithic task rendering with many different engines, maybe a separate task should be done for each template engine? Some thought can be given to providing some helper functions for creating render tasks so that users can easily create their own if they need some specific template engine.

podviaznikov:
@Deraen I also probably will use Hiccup most of the times. But my motivation was to make it easier for other people to start.
I was thinking about separate task that will use templates. But the problem is that than it should be at least 2 task: `render-template` and `collection-template`. I don't like it that much.

Also I feel like this implementation is quite flexible for extending. Because it works as before user may provide `(render :renderer)` function, that can be Hiccup rendering or anything else. What I did is to make   `(render :renderer)` optional. So the user can either provide this function (as you or me), or use included batteries for templates (provide template name and template engine).

Deraen:
The things which bothers me is that if we are going to include some template engines we also need to add dependencies for them. I would like the dependencies were added only when the user is using the template engine.

If engine is going to be set using keyword, we could map keyword to dependency before creating the pod. 

Btw. currently `collection-deps` is not used.

podviaznikov:
I like idea of mapping dependency to engine (and I agree with the argument about too much dependencies).

Thanks for pointing out that `collection-deps` wasnot used.

allentiak:
Hi, @podviaznikov,

What's the current status of this PR?

allentiak:
Hi, @podviaznikov
Are you still interested in working on this?

Issue #41, opened 2015-06-29
canonical_url


podviaznikov:
Each post should have `canonical-url` property. This can be calculated as "domain"+"post url"

podviaznikov:
Implemented in #42

PR #39, opened 2015-06-25
New Plugin: build date

podviaznikov:
Add `:build-date` to each file metadata.
Also introduce notion of global/shared metadata. Add `:build-date` to it as well.

PR #42, opened 2015-07-04
Add base-metadata and canonical-url tasks

### LINKED ISSUE #14: New plugin: meta
I want to have plugin where I can specify some global settings that would be accessible to all templates.
E.x. I'm thinking that `base_url`, `site_name`, `copyright_dates` are repetitive and can be used on any page.
Right now there os no way of specifying such options globally for all tasks.

I'm thinking creating super simple task called `meta`. Some thing like this:

```
(comp (meta {:name "Hashobject Blog" :base_url "http://blog.hashobject.com"})
          (markdown)
           (draft)
          ...)

```

Thank this meta can be merge with every file meta, or it can be accessible as `(:global-metadata (meta fileset))`.

Deraen:
Fixes #14 and #41 

**Before merging** we should decide one name for this, currently the code uses both `base-metadata` and `global-meta` and I can't decide which is better.

**Breaks** existing render functions as they are now called with two arguments, global meta and files or file. I don't see way around this as it's not possible to attach the metadata to collections.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,13 +8,15 @@ that suits your needs.
 ## Plugins
 
  - markdown - parse mardown files with YAML metadata
+ - global-metadata - read global metadata from `perun.base.edn`
  - collections - generate page that takes all posts data as parameter
  - drafts - exclude pages that have `:draft` flag
  - time-to-read - calculate time to read for each page
  - sitemap - generate sitemap for site
  - rss - generate RSS feed
  - slugs - generate slugs based on any property
- - permalinks - create permalinks for each page
+ - permalink - create permalink for each page
+ - canonical-url - create canonical-url for each page
  - gravatar - find gravatar urls using emails
  - rendering to any format - flexible rendering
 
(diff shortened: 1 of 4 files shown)


Deraen:
I just now noticed that #39 also introduces global metadata.

podviaznikov:
@Deraen this looks good to me. I also was thinking about challenge of having global meta and file meta and I think your solution is good. Passing two arguments to the render functions sounds good to me.
Another option would be to always merge global meta into the file meta. But I'm not sure I like this option.

podviaznikov:
Regarding the name: we can use `global-meta`. I choose such name for #39.

Deraen:
Merging global meta to file meta is not very useful as it would make it hard to access in collection rendering when outside of posts loop. E.g. site-title and such, though it would be possible to get the metadata from first post. But I think second parameter is much more robust.

Deraen:
Naming should be consistent now.

podviaznikov:
@Deraen agree about problem of accessing merged metadata in `collection`, `sitemap` tasks etc.

podviaznikov:
Merging!

PR #44, opened 2015-07-04
Fix render when permalink doesn't contain filename

Deraen:
- File map doesn't contain filename anymore, instead key should be used
- There is no reason why permalink should contain the filename, many will probably prefer to use only the directory name. To fix file writing in this case, if permalink ends in slash, index.html is appended to filepath.

### DIFF
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -216,7 +216,11 @@
     ((resolve '~sym) ~file)))
 
 (deftask render
-  "Render pages"
+  "Render pages.
+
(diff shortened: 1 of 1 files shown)


Deraen:
I wonder if there should be `:filepath` property which would be used if set?

maybe
`(or (:filepath file) (:permalink file) filename)`

podviaznikov:
Code looks good.
What is the use case for setting `:filepath`?

Deraen:
I don't have use cases, I was just wondering if it would be useful but I guess the permalink should be enough.

PR #47, opened 2015-07-06
update markdown plugin to insert anchors automatically

podviaznikov:
**Note:** This is not final implementation.

This PR updates Markdown plugin to automatically replace all headers.
Following transformation would happen:

``` html
<h1>Header</h1>
```

would be replaced with

``` html
<h1><a href="#header"><span></span>Header</h1>
```

This is needed to be able to share link to the particular part of the page.

TBD:

Unfortunately it wasn't possible to make it as separate plugin (since markdown parsing and html generation are bundled into one task).

So this was implemented as part of `markdown` task itself. We can make it optional.
Also we can allow customization of the output transformation.

### DIFF
--- a/build.boot
+++ b/build.boot
@@ -5,6 +5,8 @@
                   [adzerk/bootlaces "0.1.9" :scope "test"]
                   [jeluard/boot-notify "0.1.2" :scope "test"]
                   [endophile "0.1.2" :scope "test"]
+                  [enlive "1.1.5" :scope "test"]
+                  [camel-snake-kebab "0.3.2" :scope "test"]
                   [circleci/clj-yaml "0.5.3" :scope "test"]
                   [time-to-read "0.1.0" :scope "test"]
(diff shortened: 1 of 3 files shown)


Deraen:
Pegdown can do this https://github.com/sirthias/pegdown/blob/master/src/main/java/org/pegdown/Extensions.java#L106

I have already have a PR to update Endophile use latest Pegdown and I can also enchance endophile with options to set the extensions for Pegdown.

podviaznikov:
Nice, that is much better.

podviaznikov:
@Deraen maybe we should use it for now until Endophile would be pushed to Clojars?

Deraen:
New Pegdown has been released so Endophile changes are no longer blocked. I'm hoping new Endophile would be released soon.

I don't have strong preference to this being used until Endophile is updated. If you need it, sure.

Deraen:
Now that we use Pegdown directly instead of through Endophile, this works.

PR #50, opened 2015-08-26
Basic Asciidoctor plugin

nicorikken:
A basic Asciidoctor plugin, which also features the ability to generate diagrams (PlantUML) and parse basic metadata from the adoc file.

THIS PLUGIN USES A FILE-BASED CONTENT TRANSPORT RATHER THAN A METADATA WAY. It is therefore not (yet) fully compatible with the other Perun modules.

This plugin adds the Asciidoctor functionality by including a Jruby Asciidoctor implementation. The resulting content is provided to the render stage using files rather than a string in the metadata. This was done to cope with string size limitations on the metadata, which was a constraint for the use-case this plugin was designed for. This plugin can be improved by adding support for metadata-based content handing, adding an Asciidoctor attribute parser, and further separating dependencies by wrapping it in a separate pod.

### DIFF
--- a/build.boot
+++ b/build.boot
@@ -1,5 +1,5 @@
 (set-env!
-  :source-paths #{"src" "test"}
+  :source-paths #{"test"}
   :resource-paths #{"src" "resources"}
   :dependencies '[[boot/core "2.1.2" :scope "provided"]
                   [adzerk/bootlaces "0.1.9" :scope "test"]
--- a/build.boot
+++ b/build.boot
@@ -1,5 +1,5 @@
 (set-env!
-  :source-paths #{"test"}
+  :source-paths #{"src" "test"}
   :resource-paths #{"src" "resources"}
   :dependencies '[[boot/core "2.1.2" :scope "provided"]
                   [adzerk/bootlaces "0.1.9" :scope "test"]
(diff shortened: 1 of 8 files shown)


podviaznikov:
@nicorikken thanks for the PR. I need at least several days to follow up on this

podviaznikov:
@nicorikken boot has no problem any more with large metadata, so you can use regular way to pass metadata.
Do you want to update this PR with master code or you want me to take it from here?

martinklepsch:
> boot has no problem any more with large metadata, so you can use regular way to pass metadata.

@podviaznikov by that you mean using `boot.pod/send!` right? :)

podviaznikov:
@martinklepsch yes, I think so

nicorikken:
I think it's fair to say this PR is soon to be replaced by PR #100 which uses the power of AsciidoctorJ's JRuby wrapper, instead of a DIY solution. I'd still have to get the libraries like `asciidoctor-diagram` to work however.

nicorikken:
Superseded by #100 (based on Asciidoctorj instead of a raw JRuby container) and now #183 (more minimal Asciidoctorj implementation based on the more current Perun framework). Therefore can be closed.

Issue #43, opened 2015-07-04
New plugin: Atom XML


Deraen:
RSS doesn't contain the complete post content so it's quite useless, I guess Atom XML is more common these days. Luckily it's trivial to write the XML using data.xml even in the project, but I guess this is so common that we should provide an existing task for this:

``` clj
(ns blog.views.atom
  (:require [clojure.data.xml :as xml]
            [blog.dates :refer [datestr]]
            [blog.views.common :as common]))

(defn render
  [{:keys [author base-url site-title]}
   posts]
  (xml/emit-str
    (xml/sexp-as-element
      [:feed {:xmlns "http://www.w3.org/2005/Atom"}
       [:title site-title]
       [:link {:href (str base-url "atom.xml") :rel "self"}]
       [:link {:href base-url}]
       ; [:updated "fixme"]
       [:id base-url]
       [:author
        [:name (:name author)]
        [:email (:email author)]]
       (for [{:keys [canonical-url content name date-published]} (take 10 posts)]
         [:entry
          [:title name]
          [:link canonical-url]
          ; (if date-updated [:updated date-updated])
          [:content {:type "html"} (str content)]
          ])])))
```

podviaznikov:
I'm up for including this into perun as task for sure:)

podviaznikov:
I'm going to convert this code to perun plugin today.

PR #51, opened 2015-09-10
Atom Feed generation

podviaznikov:
Implementation mostly taken from #43.


podviaznikov:
Going to merge it. Can always improve it in the future

PR #52, opened 2015-09-19
allow more versatile collection rendering via group-by

martinklepsch:
Took me a while but I'm finally back at migrating my blog to perun.

I was playing with different approaches regarding pagination and this seems to be the most fruitful. 

Few thoughts:
- Shouldn't we use `ifn?` instead of `clojure.test/function` for things like `sortby` as keys could also be passed?
- maybe just fail if `:page` and `:groupby` are used to avoid potential confusion arising from not seeing warnings?
- alternative suggestions for `groupby` option docstring are welcome

Cheers

### DIFF
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -319,6 +319,7 @@
 (def ^:private +collection-defaults+
   {:out-dir "public"
    :filterer identity
+   :groupby (fn [x] "index.html")
    :sortby (fn [file] (:date-published file))
    :comparator (fn [i1 i2] (compare i2 i1))})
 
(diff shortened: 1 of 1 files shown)


Deraen:
Looks good.
- Yes, `ifn?` should be used.

podviaznikov:
Looks good to me. Should I merge it or you want to add `ifn?` in this PR?

martinklepsch:
I'll add `ifn?` calls. Any thoughts regarding fail vs warn? I'm thinking fail is most beginner-friendly.

podviaznikov:
@martinklepsch `warn` makes sense to me. Good point about being `beginner-friendly`

martinklepsch:
Well the point about beginner friendliness was pro `fail` and not warn. ;P

http://www.martinklepsch.org/ is now running on Perun btw :)

src/io/perun.clj
```diff
@@ -328,25 +328,33 @@
...
+    (cond (not (fn? (:comparator options)))
               (u/fail "collection task :comparator option should be a function\n")
-          (not (test/function? (:filterer options)))
+          (not (ifn? (:filterer options)))
               (u/fail "collection task :filterer option should be a function\n")
-          (not (test/function? (:sortby options)))
+          (and (:page options) (:groupby *opts*))
+              (u/warn "using the :page option will render any :groupby option setting effectless\n")
+          (not (ifn? (:groupby options)))
+              (u/fail "collection task :groupby option should implement IFn\n")
```
podviaznikov: please change `should implement IFn` to `should be a function` in this case.
Or make opposite change in the other cases.

I don't have any preferences for the message itself, just care about consistency

podviaznikov:
Congrats on porting your blog to perun!

I see now that we are using `fail` in all other cases. Let's use it also for your new check.

I misread both code and your comment.

martinklepsch:
amended the changes you suggested.

podviaznikov:
thanks!

martinklepsch:
:+1:

Issue #56, opened 2015-09-24
Key metadata by path instead of filename


martinklepsch:
Metadata in perun should be keyed by the path relative to the fileset instead of filenames. Plain filenames can be overriden by files with the same name.

I'm also thinking of redundantly adding the path to the metadata map which would easily allow filtering and grouping by path without having to deal with MapEntries all the time.

martinklepsch:
Also adding the `:path` to the map would allow to easily construct a perun metadata map from a list of maps.

Deraen:
:+1:

podviaznikov:
@martinklepsch your PR is merged now. Do we want to do something else in the scope of this issue? Do we want to remove `:filename` field completely from the code?

martinklepsch:
I think it's fine as is. File name might still be useful in some cases.

PR #57, opened 2015-09-25
preliminary version of new metadata structure

martinklepsch:
Partly as discussed in #56 with some extras.

This changes the format of perun metadata from being a map to a list of maps. When using `io.perun.core/set-meta` and `io.perun.core/get-meta` these maps are transformed into one map using the values of the `:path` key as keys and if two metadata sets need to be merged this mechanism can be used via `io.perun.core/merge-meta`.

Not having to handle MapEntries every time simplifies metadata transformations and functions like `map-vals` are not as useful when you need both, the information in the key and in the value of a MapEntry.

I added a `base` task that establishes some initial structure. I'm not yet convinced this is useful in some way. Probably tasks that use the added `:filename` key should just look that up through the fileset (and then also add it to the metadata).

### DIFF
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -26,6 +26,27 @@
   '[[org.pegdown/pegdown "1.6.0"]
     [circleci/clj-yaml "0.5.3"]])
 
+(deftask dump-meta
+  "Utility task to dump perun metadata via boot.util/info"
+  [r to-remove REMOVE [kw] "Keys to remove from metadata map values"]
+  (boot/with-pre-wrap fileset
+    (prn (pr-str (map #(apply dissoc % to-remove) (perun/get-meta fileset))))
+    fileset))
+
+(deftask base
+  "Adds some basic information to the perun metadata and
+   establishes metadata structure."
+  []
+  (boot/with-pre-wrap fileset
+    (let [files         (perun/get-meta fileset)
+          updated-files (map (fn [f] {:filename (.getName (boot/tmp-file f))
+                                      :path     (boot/tmp-path f)})
+                             (boot/user-files fileset))]
+      (perun/set-meta fileset updated-files))))
+
+;; See https://github.com/boot-clj/boot/issues/250
+(defn marshall-tmpfile [tmpfile] (update tmpfile :dir #(.getPath %)))
+
 (deftask markdown
   "Parse markdown files
 
@@ -37,22 +58,22 @@
         prev-meta (atom {})
         prev-fs   (atom nil)]
     (boot/with-pre-wrap fileset
-      (let [markdown-files   (->> fileset
-                                  (boot/fileset-diff @prev-fs)
-                                  boot/user-files
-                                  (boot/by-ext ["md" "markdown"])
-                                  (map #(.getPath (boot/tmp-file %))))
-            removed-files    (->> fileset
-                                  (boot/fileset-removed @prev-fs)
-                                  boot/user-files
-                                  (boot/by-ext ["md" "markdown"])
-                                  (map #(.getName (boot/tmp-file %))))
-            parsed-metadata  (pod/with-call-in @pod
-                               (io.perun.markdown/parse-markdown ~markdown-files
-                                                                 ~options))
-            initial-metadata @prev-meta
-            final-metadata   (merge initial-metadata parsed-metadata)
-            final-metadata   (apply dissoc final-metadata removed-files)
+      (let [md-files (->> fileset
+                          (boot/fileset-diff @prev-fs)
+                          boot/user-files
+                          (boot/by-ext ["md" "markdown"])
+                          (map marshall-tmpfile))
+            removed? (->> fileset
+                          (boot/fileset-removed @prev-fs)
+                          boot/user-files
+                          (boot/by-ext ["md" "markdown"])
+                          (map #(:path (boot/tmp-file %)))
+                          set)
+            md-meta  (pod/with-call-in @pod
+                       (io.perun.markdown/parse-markdown ~md-files ~options))
+            initial-metadata (perun/merge-meta (perun/get-meta fileset) @prev-meta)
+            final-metadata   (perun/merge-meta initial-metadata md-meta)
+            final-metadata   (remove #(-> % :path removed?) final-metadata)
             fs-with-meta     (perun/set-meta fileset final-metadata)]
         (reset! prev-fs fileset)
         (reset! prev-meta final-metadata)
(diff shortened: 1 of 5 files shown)


src/io/perun.clj
```diff
@@ -101,20 +122,21 @@
...
 (deftask draft
   "Exclude draft files"
   []
   (boot/with-pre-wrap fileset
     (let [files         (perun/get-meta fileset)
-          updated-files (perun/filter-vals #(not (true? (:draft %))) files)
+          updated-files (filter #(not (true? (:draft %))) files)
           fs-with-meta  (perun/set-meta fileset updated-files)]
       (u/info "Remove draft files. Remaining %s files\n" (count updated-files))
       fs-with-meta)))
```
martinklepsch, line 143: I think there is no need for a separate draft task. Tasks that render stuff should always have a `filterer` option allowing this kind of behaviour (and potentially covering more complex scenarios like tags or categories).

src/io/perun.clj
```diff
@@ -284,11 +295,18 @@
...
   {:pre [(symbol? sym) (namespace sym)]}
-  (pod/with-eval-in pod
-    (require '~(symbol (namespace sym)))
-    ((resolve '~sym) ~global-meta ~file)))
+  ;; Ensure passed seqs are vectors, otherwise lists/array-seqs etc
+  ;; wrapped in parentheses will be interpreted as function calls
+  (let [m (if (sequential? file-meta) (vec file-meta) file-meta)]
+    (pod/with-eval-in pod
+      (require '~(symbol (namespace sym)))
+      ((resolve '~sym) ~global-meta ~m))))
```
martinklepsch, line 306: This got me a couple of times when passing filtered lists to the `render-in-pod` function.

martinklepsch:
Makes things really [wieldy](https://www.youtube.com/watch?v=2YmmfS5YTQ0). :+1:

src/io/perun.clj
podviaznikov, line 143: `draft` task makes and inpact on all downstream tasks not only `render`. E.x. we don't want to include draft posts into the sitemaps etc. It's something that should be done on the top of the tasks chain and not in the bottom.

Maybe we can create more general `filter` task instead of `draft` but that is another question

src/io/perun.clj
```diff
@@ -26,6 +26,27 @@
   '[[org.pegdown/pegdown "1.6.0"]
     [circleci/clj-yaml "0.5.3"]])
 
+(deftask dump-meta
```
podviaznikov, line 29: I like this task. It's just for debugging right? So user can see what is in meta between each tasks

src/io/perun.clj
```diff
@@ -26,6 +26,27 @@
...
+  "Utility task to dump perun metadata via boot.util/info"
+  [r to-remove REMOVE [kw] "Keys to remove from metadata map values"]
+  (boot/with-pre-wrap fileset
+    (prn (pr-str (map #(apply dissoc % to-remove) (perun/get-meta fileset))))
+    fileset))
+
+(deftask base
```
podviaznikov, line 37: Do I need to call this task before all other tasks?

src/io/perun.clj
```diff
@@ -312,15 +333,15 @@
...
+            (u/dbug " -> %s\n" page-filepath)
             (perun/create-file tmp page-filepath html)))
-        (u/info "Render all pages\n")
         (commit fileset tmp)))))
 
 (def ^:private +collection-defaults+
   {:out-dir "public"
    :filterer identity
-   :groupby (fn [x] "index.html")
+   :groupby (fn [data] "index.html")
```
podviaznikov, line 345: :+1:

podviaznikov:
I like those changes. Really great job. Now it's easier to write tasks and read/update metadata

src/io/perun.clj
martinklepsch, line 143: I'm thinking it would be more useful to encourage people to use something like `(merge-env! :resource-paths '["_drafts"])` for this. Less code to maintain + some education about Boot :)

src/io/perun.clj
martinklepsch, line 29: Yes it is. It's probably not ideal yet but in general it probably wont hurt to provide something along those lines. :) I just pushed another commit with another `map-fn` option that turned out to be more useful.

src/io/perun.clj
```diff
@@ -26,6 +26,29 @@
   '[[org.pegdown/pegdown "1.6.0"]
     [circleci/clj-yaml "0.5.3"]])
 
+(deftask dump-meta
+  "Utility task to dump perun metadata via boot.util/info"
+  [r to-remove REMOVE [kw] "Keys to remove from metadata map values"
```
podviaznikov: I guess `r` option is not used anymore?

podviaznikov: I like this task. Should be useful. :+1:

podviaznikov:
I like this PR, going to merge. Only question: is it required to call `base` task?

src/io/perun.clj
podviaznikov, line 143: Still not 100% convinced. Let's create ticket with discussion for it.

src/io/perun.clj
martinklepsch, line 37: From the PR text:

> I added a `base` task that establishes some initial structure. I'm not yet convinced this is useful in some way. Probably tasks that use the added `:filename` key should just look that up through the fileset (and then also add it to the metadata).

I'm not sure if there's a need for this kind of task that provides an initial set of things to work with but with the present code this has been effectively handled by the markdown task based on the assumption that it comes first. (Or at least that's how I understood the code.) 

In a way the markdown task still does it so I'm not sure if it's 100% necessary. Input by you and others definitely appreciated :)

martinklepsch:
@podviaznikov regarding `base` task: https://github.com/hashobject/perun/pull/57#discussion_r40943438

podviaznikov:
@Deraen are you good with this PR? Any comments?

Deraen:
:+1:

podviaznikov:
Merged, thanks @martinklepsch

martinklepsch:
@podviaznikov maybe let's also cut a new version and add a minimal changelog?

podviaznikov:
Hm, got `clojure.lang.ExceptionInfo: Unable to resolve symbol: update in this context
    data: {:file "io/perun.clj", :line 49}
java.lang.RuntimeException: Unable to resolve symbol: update in this context` when tried to run `boot build` on master

Deraen:
Update requires Clojure 1.7.

podviaznikov:
Works now. Should I add `boot.properties` file then to this repo?

martinklepsch:
Probably a good idea but `boot.properties` will not cascade to projects
depending on perun so maybe an adding a note to the readme would be good
too. :)

Sorry for the breakage. Theoretically 1.6 compat is possible but I guess we
don't care?

Deraen:
I think 1.7 should be default already if one is running up-to-date Boot (2.2.0+).

podviaznikov:
Yeah, no problem at all. I don't think we care about older versions of Clojure.

Deraen:
Boot `boot.properties` and mention on readme would still be good idea :)

podviaznikov:
I see, going to update my boot, add boot.properties and mention it in readme.

Deraen:
Updating boot dependency on the build.boot is also good idea.

podviaznikov:
I think it's done now: added boot.properties, updated boot dependency in the build.boot, updated readme.

PR #59, opened 2015-10-09
Separate library & example Boot files

podviaznikov:
Instead of having single boot file for library and example we will have example code isolated into `example` dir.


podviaznikov:
@Deraen, @martinklepsch I'm having small problem separating example code.

When I run `boot build` inside `example` dir everything works fine when `gravatar` plugin disabled.
But when it's enabled I'm getting following error:

```
Added slugs to 1 files
             clojure.lang.ExceptionInfo: java.io.FileNotFoundException: Could not locate gravatar__init.class or gravatar.clj on classpath., compiling:(io/perun/gravatar.clj:1:1)
    data: {:file
           "/var/folders/7d/9c32qg7x2_z9slnznbhq2dg80000gn/T/boot.user3280952706267273543.clj",
           :line 19}
java.util.concurrent.ExecutionException: java.io.FileNotFoundException: Could not locate gravatar__init.class or gravatar.clj on classpath., compiling:(io/perun/gravatar.clj:1:1)
clojure.lang.Compiler$CompilerException: java.io.FileNotFoundException: Could not locate gravatar__init.class or gravatar.clj on classpath., compiling:(io/perun/gravatar.clj:1:1)
          java.io.FileNotFoundException: Could not locate gravatar__init.class or gravatar.clj on classpath.
                                      ...
                     clojure.core/load/fn      core.clj: 5866
```

Any quick ideas why that might happen?

Deraen:
:+1: for this change

https://github.com/hashobject/perun/blob/master/src/io/perun.clj#L125 Gravatar pod is using wrong deps.

podviaznikov:
Thanks @Deraen! Mistery solved

PR #67, opened 2015-10-13
Revert "fix filename calculation"

martinklepsch:
This reverts commit 1486803407a425f7b627161971686823753c94e9.

### DIFF
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -35,13 +35,9 @@
     fileset))
 
 (defn add-filedata [f]
-  (let [tmpfile (boot/tmp-file f)]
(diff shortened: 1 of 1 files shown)


podviaznikov:
Ahh, I tested this with 1 example site and 1 real site and still broke it. Let me quickly test this revert to see  if it breaks something else

podviaznikov:
Everything works fine. Now I'm confused why I did that change. I don't see code that relies on it.

podviaznikov:
thanks for fixing! @martinklepsch :+1:

martinklepsch:
Perfect. Thanks! :)

PR #69, opened 2015-10-17
inject-scripts tasks

podviaznikov:
Inspired by https://github.com/tailrecursion/hoplon.io/blob/master/build.boot#L40
and https://github.com/Deraen/boot-hyphenate/


Deraen:
I think this should be contrib task?

podviaznikov:
Now it's in `contrib`

Issue #53, opened 2015-09-19
Use global metadata for feed options


Deraen:
I use global metadata to set the title and url of the site so it would be very useful if I could use that for feeds also. As the global metadata schema is not defined this is not very straightforward.

Alternatives:
1. Define some global metadata keys which will be used by built-in tasks if set
2. Add options to the tasks to set which keys to use

podviaznikov:
I think we should defined some global metadata keys. And use them in built in tasks if they are present and local task option was not provided.

Deraen:
I started a document listing all the metadata fields: https://github.com/hashobject/perun/blob/spec/SPEC.md

I think few changes would be in order for global and post metadata to be consistent:
- General author schema used by global and post metadata
  - Does yaml metadata support sub keys, e.g. `[:author :name]`?
- `build-date`, `date-published`  should be consistent
- ~~build-date could be global key~~ build-date is both global and post metadata. Would it be enough it it was only global?

podviaznikov:
I like that we have formal listing of all metadata fields.
1. I don't know about YAML metadata support for nested keys but I'll check it tomorrow.
2. I think we can ged rid of `build-date` key and make `build-date` plugin to set `date-published`. I like `date-published` because it's consistent with `date-created` and `date-updated`

Deraen:
Build date is different as published. It's the date when Perun was run.

podviaznikov:
so what do you mean by ?

> build-date, date-published should be consistent

podviaznikov:
@Deraen clj-yaml supports nested keys: https://github.com/lancepantz/clj-yaml/blob/2dc440aa2016165aabb16408d17f523dee403c8f/test/clj_yaml/core_test.clj#L62

podviaznikov:
@Deraen I think part of this task is done now.
I don't understand what normalization/changes should be done to `build-date` and `date-published`.
Also not 100% sure of using nesting for `:author`. What are the benefits of having `:author :name` instead of `:author-name`?

Deraen:
Not sure about nesting myself.

Perhaps `build-date` should be renamed to `date-build`?

podviaznikov:
I see now, `date-build` makes more sense. Agree with suggestion and will do it.
Cool, lets not do nesting for now, until we have good reasons.

PR #70, opened 2015-10-17
Use global meta for feeds

podviaznikov:
Implement part of the issues outlined https://github.com/hashobject/perun/issues/53


Deraen:
Spec should also be updated for these changes.

podviaznikov:
agree, going to update. Are you fine with those changes? Or you have better suggestions?

Deraen:
Looks good.

PR #74, opened 2015-11-05
Pass all pages to render

podviaznikov:
So I changed signature of the render function that user needs to implement.
Instead of passing `global-meta` and `file-meta` we would have `global-meta`, `files-meta` and `file-meta`.
This is important because sometimes on the page you want to render navigation and to do that you need information about all files.
PR also includes minor documentation changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,9 +157,9 @@ See documentation for each task to find all supported options for each plugin.
 (require '[hashobject.boot-s3 :refer :all])
 (require '[jeluard.boot-notify :refer [notify]])
 
-(defn renderer [global data] (:name data))
+(defn renderer [{global :meta posts :entries post :entry}] (:name post))
 
-(defn index-renderer [global files]
+(defn index-renderer [{global :meta files :entries}]
   (let [names (map :name files)]
     (clojure.string/join "\n" names)))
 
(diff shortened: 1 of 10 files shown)


podviaznikov:
@martinklepsch, @Deraen does this PR makes sense? It changes signature of user's render function, so I want to make sure it's a good change.

martinklepsch:
There are a bunch of other changes in this PR that make reviewing that particular change a little harder. I suggest in the future splitting style changes and "feature code". :smiley: 

This change would kind of remove the need for a collection task I guess?

I assume you needed this for something, would you mind sharing the example where this came in useful?

podviaznikov:
@martinklepsch 
1. agree with your comment on the clean PRs.
2. it seems that collection task is still needed. even in this case they are different. render has all the same arguments as collection but it also has current post to be rendered.
3. yes, I need it for the case which I described above. While rendering post I also want to render navigation with the links to other post. Basically this page: http://jack.ventures/2014/hongkong/. When you can see links to other posts on the left

Deraen:
How about passing the data in as map instead of positional args? This way we wouldn't need to change the arity each time we add more data to render function?

All render functions have one map argument and the map would always two entries, `meta` and `posts` and for single render task it would contain `post` also?

podviaznikov:
@Deraen I like that. Going to change it to one argument. Thanks for the feedback!

podviaznikov:
Updated renderers to accepts only one argument.

Note: I put global meta under `:meta`, current post under `:entry` and all posts under `:entries`.

I'm thinking about using something more generic than `:post` since perun can be used not only for blog generation. Let me know what do you think

PR #76, opened 2015-11-30
Better logs

podviaznikov:
1. Use colors for logs and better formatting
2. Log message include task name
3. Each task prints message only once
4. Each task can can additional message visible with the `—verbose` flag


podviaznikov:
1__boot_dev__java__and_hashobject_perun

martinklepsch:
Why the angle brackets around the messages? I think I'd suggest turning the
task name green and leaving the rest white.

podviaznikov:
@martinklepsch thanks for the feedback. Will try you suggestion

PR #79, opened 2016-01-16
Images task

podviaznikov:
Extracted this plugin from my other experimental project.
Should be used with the following structure:
- create folder for each page
- put markdown file and images inside each folder
- after using markdown task and images task images data would be available under `:images` key on the post metadata

Not fully tested it. Need to create proper example

### DIFF
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -91,7 +91,7 @@
           pod (create-pod images-resize-deps)
           files (->> fileset
                      boot/user-files
-                     (boot/by-ext ["png" "jpeg" "jpg"])
+                     (boot/by-ext perun/images-exts)
                      (map add-filedata))
           updated-files (pod/with-call-in @pod
                          (io.perun.contrib.images-resize/images-resize ~(.getPath tmp) ~files ~options))]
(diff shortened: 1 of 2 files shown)


nicorikken:
Would love :heart: this for the Asciidoctor plugin. There the `asciidoctor-diagram` extension can render images from inline source. Rendered SVG's can optionally be included in the HTML (powerful feature!!), but otherwise external images (`.png`, `.svg`) will be created. The filename can be specified by the user, but I'm not sure if this can extend to outside of the file's directory.

nicorikken:
I'd like a sample project or concrete example for this PR, because the Asciidoctor could benefit from this as well. I can probably scan the AST of the document to find the linked images, as well as the generated images, and then add them to the metadata. I guess I can benefit from a better understanding on the file management as a whole.

nicorikken:
@podviaznikov How does this relate to the `content-task` framework? Perhaps the image handling in the `content-task` can be improved instead of a separate task?

allentiak:
Hi, @podviaznikov,

What's the current status of this PR?

PR #81, opened 2016-01-17
Move any special handling of :content key into options

martinklepsch:


src/io/perun.clj
```diff
@@ -333,28 +335,26 @@
...
         tmp (boot/tmp-dir!)]
     (boot/with-pre-wrap fileset
       (let [global-meta   (perun/get-global-meta fileset)
             options       (merge +atom-defaults+ global-meta *opts*)
-            files         (perun/get-meta fileset)
-            content-files (filter :content files)
-            filtered-files (filter (:filterer options) content-files)]
+            files         (filter (:filterer options) (perun/get-meta fileset))]
         (pod/with-call-in @pod
-          (io.perun.atom/generate-atom ~(.getPath tmp) ~filtered-files ~(dissoc options :filterer)))
```
martinklepsch, line 357: @Deraen is the `(dissoc options :filterer)` bit important?

Deraen, line 357: Yes. A function can not be passed into pod.

martinklepsch:
@Deraen @podviaznikov this PR modifies all tasks that usually only act on files that have a `:content` key. It adds a `:filterer` option as done with other tasks. `:content` is the default value for this option intending not to break to much stuff. Maybe we should break things, not sure.

Deraen:
Hhmh. Maybe related, I find it strange that collection task adds files to metadata. What is the use case for that?

src/io/perun.clj
martinklepsch, line 357: Right, thanks! I thought printing works but it breaks when reading :)

podviaznikov:
@Deraen collection adds files to metadata so that sitemap can access it

Deraen:
Oh right, I guess that makes sense.

What's the best way to filter those files out? Is it already done by default by some tasks? Is this documented somewhere :D?

Maybe SPEC.md would be good place to document this also.

podviaznikov:
@Deraen I don't think there is a way now. I think I was thinking before about `original: true` flag. Another idea is to have `created-by-task` field that would be set by task that put the file into metadata.

podviaznikov:
Code looks great to me. I like it that it's not breaking existing code. Let me quickly test it before merge

src/io/perun.clj
```diff
@@ -284,21 +284,22 @@
 
 (def ^:private +sitemap-defaults+
   {:filename "sitemap.xml"
+   :filterer :content
    :target "public"})
 
 (deftask sitemap
   "Generate sitemap"
   [f filename FILENAME str "generated sitemap filename"
+   f filterer FILTER   code "filter function"
```
podviaznikov: conflict of short name between `filterer` and `filename`

src/io/perun.clj
```diff
@@ -307,11 +308,13 @@
 
 (def ^:private +rss-defaults+
   {:filename "feed.rss"
+   :filterer :content
    :target "public"})
 
 (deftask rss
   "Generate RSS feed"
   [f filename    FILENAME    str "generated RSS feed filename"
+   f filterer    FILTER     code "filter function"
```
podviaznikov: again conflict

src/io/perun.clj
podviaznikov: maybe we should use `_` everywhere for the `filterer` as below

podviaznikov:
Looks good to me. Tested with examples/blog

PR #84, opened 2016-01-18
improve docstrings for render and collection tasks

martinklepsch:


podviaznikov:
:+1: Great job. Much better this way

PR #86, opened 2016-01-19
consolidate out-dir/target options; closes #83

### LINKED ISSUE #83: Consolidate `target` & `out-dir` options
Both are used but it's probably best to stay consistent

martinklepsch:


podviaznikov:
:+1:

PR #87, opened 2016-01-22
File-type

podviaznikov:
Work in progress.
Add `file-type` task that will parse `:mime-type` and `:file-type` for each file.
Can be used for easy filtering files based on those keys later.


podviaznikov:
new task is ready for review. Feedback appreciated

Deraen:
:+1:

PR #90, opened 2016-02-11
Fix feed task options

AdamFrey:
At some point, the option names passed to `rss` and `atom-feed` tasks
were changed, but the feed building code wasn't updated to expect the
new option names.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,8 +81,8 @@ But if you want to make permalinks, generate sitemap and rss feed, hide unfinish
         (permalink)
         (render :renderer renderer)
         (sitemap :filename "sitemap.xml")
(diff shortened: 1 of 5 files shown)


podviaznikov:
Thank you @AdamFrey!

Deraen:
This fixed this backwards. The correct fix would have been to fix feed generation code. `:site-title` and `:base-url` are set by the [spec](https://github.com/hashobject/perun/blob/master/SPEC.md).

podviaznikov:
Oh, that was y bad. Didn't double check with the spec

Deraen:
@podviaznikov Fixed this and some other Atom problems here: https://github.com/hashobject/perun/tree/atom-fixes will try to finish soon.

podviaznikov:
@Deraen your PR is looking like improvement already

PR #92, opened 2016-02-23
Atom fixes

Deraen:
TODO:
- [x] Validate `base-url`. It is now important that this does end in `/`.
- [x] Validate that UUID is set for Atom entries, maybe add suggestion about adding uuid if not found?
- [ ] Test what problems moving `/` from permalink to base-url has. Might break `a` using permalink to point to other directories?

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,8 +82,8 @@ But if you want to make permalinks, generate sitemap and rss feed, hide unfinish
         (permalink)
         (render :renderer renderer)
         (sitemap :filename "sitemap.xml")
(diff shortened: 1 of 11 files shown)


Deraen:
Ready for merge now.

This has a few breaking changes so would be good if someone has a time to test this.

podviaznikov:
so if I use atom I need to specify manually `uuid` for each post?

Deraen:
Yes. I don't see other reliable way to ensure that it doesn't change.

Even the slug can be changed if the file is renamed.

PR #98, opened 2016-08-10
change posts primary title field from :name to :title

martinklepsch:
Brings things more in line with existing static site generators like jekyll.


nicorikken:
When will this be merged? This could save the Asciidoctor plugin a transformation from :title to :name for the time being 😛

podviaznikov:
This would be a breaking change but makes sense to me. Didn't see it before

podviaznikov:
@Deraen do you have any opinion on this?

Deraen:
Fine by me.

PR #102, opened 2016-09-28
Insert inline js at body end instead of header end

edannenberg:
It's best practise for inline js, minimizes render blocking.

### DIFF
--- a/src/io/perun/contrib/inject_scripts.clj
+++ b/src/io/perun/contrib/inject_scripts.clj
@@ -4,7 +4,7 @@
 
 (defn inject [html scripts]
   (reduce
-    #(.replaceFirst %1 "</head>" (format "<script>%s</script></head>" %2))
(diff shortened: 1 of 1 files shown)


podviaznikov:
I think that makes sense. But also it seems that user should be able to determine where to insert js?
Maybe add additional parameter and by default insert at the end of `body`?

edannenberg:
I was thinking about that, but couldn't really find a reason where you want your inline js in the header. Wouldn't hurt though I guess.

martinklepsch:
I'd suggest not adding an option unless someone can come up why that option is necessary. The more knobs we add the harder will it be for users to see which ones are important/relevant to them.

podviaznikov:
Merging then. Arguments make sense.
Thank you @edannenberg for PR and @martinklepsch for input

PR #104, opened 2016-10-18
Use tools namespace to reload changed clojure ns instead of pod pools

Deraen:
# Reasoning
- Current pod pool implementation is slow, after change to any clj file, render takes 15 seconds
- Each task has it's own pod pool -> 2 render tasks and 3 collection tasks -> 5 pod pools.
  Each pod pool is refreshed when a clj file changes, this is done sequentially so each render task adds to the execution time.
- If only single pod pool would be used recompilation would probably take only 3 seconds (but even this is too slow for me).
# With tools namespace
- Single pod for all render tasks (not sure if this can cause problems?)
- Any render task uses c.t.n to reload change clojure ns (and unload deleted ns etc.), usually no-op for all but the first task.
- Recompilation in 0.2 sec

### DIFF
--- a/build.boot
+++ b/build.boot
@@ -11,7 +11,8 @@
                   [gravatar "1.1.1" :scope "test"]
                   [clj-time "0.12.0" :scope "test"]
                   [mvxcvi/puget "1.0.0" :scope "test"]
-                  [com.novemberain/pantomime "2.8.0" :scope "test"]])
+                  [com.novemberain/pantomime "2.8.0" :scope "test"]
+                  [org.clojure/tools.namespace "0.3.0-alpha3" :scope "test"]])
 
 (require '[adzerk.bootlaces :refer :all])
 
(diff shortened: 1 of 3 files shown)


martinklepsch:
This looks exciting, I've been a little but frustrated with slow compile
cycles lately.

Will give it a test run later today 👍

martinklepsch:
So just went ahead and gave it a try now. Previously had 15s compile cycles now more like 2s. 👏 

Looking at the code I'm thinking maybe we should also change this:

```
(defn- create-pod
  [deps]
  (future (create-pod' deps)))
```

to this??

```
(defn- create-pod
  [deps]
  (delay (create-pod' deps)))
```

podviaznikov:
+1. Same for me. Created new site with perun and it was too slow to compile. This sounds great

martinklepsch:
@Deraen pushed a change, `update!` fn wasn't in the right place for the `collection` task I believe.  You only notice when you use only the `collection` task.

martinklepsch:
Another thing that I ran into: when you have something in `src` that require's `boot.core` (i.e. some task) this change breaks the reloading. I'm not sure if there's a way to mitigate that so mostly logging it here to keep track.

(BTW, I tried adding boot as an explicit dependency but that didn't fix the issue)

Deraen:
The latest change fixes that. Now only namespaces that are already loaded, e.g. transitively by namespace required by render task, are reloaded.

Issue #105, opened 2016-10-21
Feature: pass pre-parse function to markdown task


edannenberg:
Use case:

Our editors want to set, for example, the `rel`  or `target` attribute for links while using markdown syntax. Inspired by Jekyll I extended the markdown syntax for our frontend editor like so:

```
[foo](bar.html){:rel "nofollow" :blank? true :title "foo bar"}
```

The tag is handled (replaced with final html) via re-seq/read-string before the frontend markdown parser does it's magic for the preview. Now I would like to be able to pass a `:pre-parse-fn` to the perun markdown task to do the same for the backend part.

Took a stab at the feature but as the markdown parsing happens in it's own pod this seems impossible?
Ideally the passed function would run after https://github.com/hashobject/perun/blob/master/src/io/perun/markdown.clj#L97

I could probably create a separate task that does the pre processing before the markdown task but would like to hear your opinions first. Happy to create a PR if this is somehow feasible.

Deraen:
You can pass symbol to pods and then resolve the symbol to function in the pod, similar to: https://github.com/hashobject/perun/blob/master/src/io/perun.clj#L387-L388

In this case you pretty much need to def the function in some namespace instead of writing it inline in build.boot. Also to reflect changes to the function you would need to use tools.namespace or such to reload changed namespaces.

edannenberg:
Thanks @Deraen, I've put up a quick draft PR. Does the job for my needs, hopefully I didn't break anything else in the process. :p

PR #106, opened 2016-10-24
Add preprocessor option for markdown task

edannenberg:
See #105 for details. Not sure if it would have been better to put `:preprocessor` inside the `:options` map instead.

### DIFF
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -118,10 +118,16 @@
   '[[org.pegdown/pegdown "1.6.0"]
     [circleci/clj-yaml "0.5.5"]])
 
+
 (def ^:private +markdown-defaults+
   {:meta {:original true
           :include-rss true
-          :include-atom true}})
+          :include-atom true
+          :preprocessor 'clojure.core/identity}})
+
+(defn- assert-preprocessor [sym]
+  (assert (and (symbol? sym) (namespace sym))
+  "Markdown :preprocessor must be a fully qualified symbol, i.e. 'my.ns/fun"))
 
 (deftask markdown
   "Parse markdown files
@@ -130,10 +136,13 @@
    and add a `:content` key to their metadata containing the
    HTML resulting from processing markdown file's content"
   [m meta    META edn "metadata to set on each entry; keys here will be overridden by metadata in each file"
-   o options OPTS edn "options to be passed to the markdown parser"]
+   o options OPTS edn "options to be passed to the markdown parser"
+   p preprocessor PREPROCESSOR sym "process markdown before it is parsed (fully qualified symbol which resolves to a function)"]
   (let [pod       (create-pod markdown-deps)
         prev-meta (atom {})
-        prev-fs   (atom nil)]
+        prev-fs   (atom nil)
+        merged-def (merge +markdown-defaults+ *opts*)]
+    (assert-preprocessor (:preprocessor merged-def))
     (boot/with-pre-wrap fileset
       (let [options  (merge +markdown-defaults+ *opts*)
             md-files (->> (boot/fileset-diff @prev-fs fileset :hash)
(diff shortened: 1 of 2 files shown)


MartyGentillon:
Does the preprocessor reload correctly when changed without restarting boot?

(I am noticing the pod pool being used for this in the render task, and wondering if you need something similar.  Alternately, using tools namespace as in https://github.com/hashobject/perun/pull/104)

MartyGentillon:
One other sidenote: overall this feels like the wrong approach, a result of the underlying design flaw in markdown processing (and perhaps the library overall).  Ideally, tasks should be built by composing simpler subtasks.  the markdown task should be defined something like:

``` clojure
(deftask markdown 
  (process-yaml-frontmatter)
  (parse-markdown))
```

or perhaps even:

``` clojure
(deftask markdown
  (slurp-file-contents)
  (process-yaml-frontmatter)
  (parse-markdown))
```

If it were set up this way, then there would be no need for Perun to know about markdown preprocessing.  You could simply add a preprocess-markdown task to a my-markdown task.

A couple other benefits to this approach:

1) The process-yaml-frontmatter functionality could be more easily shared between markdown, ascidoc, and html-fragment tasks.
2) Other transformations of the html output become more tenable.  For instance, someone could add a link checker that makes sure all the links in the website are valid.

The primary design flaw which prevents this is that the add-filedata function overwrites any processing done by any previous tasks.  If it instead only added data for missing files, we would be able to approach the problem this way.

None of this is to say that this pull request should be thrown away.  It's possible that this is a good short term solution for the problem being encountered.  Just that the alternate approach is worthy of consideration.

edannenberg:
> Does the preprocessor reload correctly when changed without restarting boot?

Nope, it's just a quick 90% solution at this point that removes an impediment for me. Will take a look at reloading once I got a bit more time on my hands.

As I'm still fairly new to Perun I'll let others comment on your other points, once some consensus is reached on how this should be implemented I'll update the PR.

podviaznikov:
@MartyGentillon I would love to have possibility to have better `markdown` task myself. I do think it should offer more customization.
E.x. I have personal use case for post-processor part of markdown task.
My case is following:
1) convert markdown to html.
2) add tachyons css classes http://tachyons.io/ to the appropriate HTML tags.

Right now I'm doing it in the following way:

```
(defn transform-html [html]
  (-> html
    (cue/replace "<h2>" "<h2 class=\"f5\">")
    (cue/replace "<h3>" "<h3 class=\"f6\">")
    (cue/replace "<strong>" "<strong class=\"b\">")
    (cue/replace "<em>" "<em class=\"i\">")
    (cue/replace "<a href" "<a class=\"link\" href")
    (cue/replace "<blockquote>" "<blockquote class=\"athelas ml0 mt0 pl4 black-90 bl bw2 b--blue\">")
    ))

(defn transform-file-content [file]
  (let [content (:content file)
        new-content (transform-html content)]
    (assoc file :content new-content)))
```

But it would be probably nice if I can just register post-processor with `markdown` task.

However I'm not sure how that should look like. Maybe it should look exactly like you suggest. Or maybe some other way is possible.

Hard for me to say at this point. Would love to hear input from others.

Deraen:
Customizing markdown (or other tasks) would be much easier if we were to run these steps (or at least the custom functions) in the "core" pod. Then there would no need for pod pools or multiple c.t.n refreshes, single c.t.n refresh in pipeline could reload all loaded namespaces.

This will need some thinking, but perhaps the benefits from pods are not that great with Perun.

bhagany:
The commit that closed this PR (and #105) makes `markdown` a composition of two simpler tasks, as suggested above.  You should be able to make your preprocessor function into a standalone task, and then compose it with the `yaml-metadata` and `markdown*` tasks to achieve the result you need.  If you run into issues that make this difficult, let us know.

PR #114, opened 2016-12-01
use explicit :include-atom and :include-rss flags for inclusion in feeds

bhagany:
Following @MartyGentillon, here: https://github.com/hashobject/perun/pull/109#issuecomment-263157303.  Intent is more clear, and this change or something like it is a prerequisite for fulfilling the goals of #109.

### DIFF
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -316,7 +316,7 @@
 
 (def ^:private +rss-defaults+
   {:filename "feed.rss"
-   :filterer :content
+   :filterer :include-rss
    :out-dir "public"})
 
 (deftask rss
@@ -344,7 +344,7 @@
 
 (def ^:private +atom-defaults+
   {:filename "atom.xml"
-   :filterer :content
+   :filterer :include-atom
    :out-dir "public"})
 
 (deftask atom-feed
--- a/src/io/perun/markdown.clj
+++ b/src/io/perun/markdown.clj
@@ -76,14 +76,19 @@
         :else y))
     x))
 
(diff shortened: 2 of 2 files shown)


martinklepsch:
Two questions:
- Can you think of situations where you'd want to render one atom-feed and another rss-feed with different items?
- This somehow couples the `content` task to the implementation of the RSS and Atom tasks as these know about those keys and read from them. Maybe the `default-meta` map should be supplied as an option and thus be extendable to the users needs? (We could still have some sensible defaults ofc)

bhagany:
- No, I just went for maximum flexibility.  I can't even think of a situation where you're using both `atom-feed` and `rss`.  In any case, this implementation is just as coupled as before, it's just that in this case the coupling goes the opposite way.  (in master, `atom-feed` and `rss` know about `markdown`, in this branch, `markdown` knows about `atom-feed` and `rss`)
 - Yes!  I have a pull request planned to modify the meta in exactly the way you describe.  In the meantime, this implementation allows for overriding the default by putting it in the yaml in markdown files, or the original method of changing the `:filterer` on `atom-feed` and `rss`.

martinklepsch:
Ok sounds good to me then 👍 

I'll just go ahead and merge @Deraen @podviaznikov

bhagany:
Great, thanks!

podviaznikov:
This looks good to me. Great change.
We only need to update our Spec file https://github.com/hashobject/perun/blob/master/SPEC.md with those new properties.

PR #116, opened 2016-12-02
Add a `:meta` task option to `markdown`, `render`, and `collection`

bhagany:
Allows more control of entry metadata in the case of `markdown`, and any
control at all in the cases of `render` and `collection`, following @MartyGentillon and @martinklepsch, in https://github.com/hashobject/perun/pull/109#issuecomment-263201298 and https://github.com/hashobject/perun/pull/114#issuecomment-264083450, respectively.

Having `:meta` in `render` isn't terribly useful now, but in the spirit of #109, it would be consistent to have this option on any task that deals with `:content`.

### DIFF
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -107,18 +107,25 @@
   '[[org.pegdown/pegdown "1.6.0"]
     [circleci/clj-yaml "0.5.5"]])
 
+(def ^:private +markdown-defaults+
+  {:meta {:original true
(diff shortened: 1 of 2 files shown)


podviaznikov (approved):
I like having those smaller PRs that implement one feature

podviaznikov:
Small conflict

bhagany:
and fixed

PR #117, opened 2016-12-02
Add `:io.perun/trace` to file metadata

bhagany:
Each built-in task that modifies file metadata conjes a keyword onto the `:perun/trace` vector. This will aid in debugging, and possibly have applications in tasks.

Following @MartyGentillon here: https://github.com/hashobject/perun/pull/113#issuecomment-263158233.  Two notes:

 - I went with namespaced keywords.  It felt right, and I was more comfortable breaking with the existing style with `:perun/trace` than with `:include-rss` and `:include-atom`, because `:perun/trace` feels like a different kind of metadata.  For the values in the vector, I went with `:perun/` namespaced keywords as well, because 3rd party code may want to hook into the `trace` functionality.  I still think that namespaced keywords are a good idea pretty much everywhere in perun, fwiw.
 - Some tasks, most notably `render`, don't use `trace`, because they don't currently modify metadata.  If #116 gets merged before this one, I'll modify this PR to `trace` `render` as well.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -77,3 +77,6 @@ All posts have a filename which is used as a key to identify the post.
 - **:include-rss**
     - Set by: *markdown*, true by default
     - Used by: *rss*
+- **:io.perun/trace**
+    - Conjed onto by every task that modifies metadata
+    - Serves as a record of tasks to touch a file
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -41,6 +41,8 @@
     fileset))
 
 (defn trace
+  "Helper function, conj `kw` onto the `:io.perun/trace` metadata
+  key of each entry in `entries`"
   [kw entries]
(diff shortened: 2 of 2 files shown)


martinklepsch:
> I still think that namespaced keywords are a good idea pretty much everywhere in perun, fwiw.

👍 

One suggestion would be `:io.perun` instead of `:perun`. Just to comply with the current namespacing scheme...

podviaznikov (approved):
Great idea. I like it

bhagany:
> One suggestion would be :io.perun instead of :perun. Just to comply with the current namespacing scheme...

I was going to make a case for the shorter thing, but `:io.perun` makes aliasing easier, so you win :).  It's done.

podviaznikov:
Small conflict, will merge when resolved

bhagany:
Okay, fixed.

PR #120, opened 2016-12-05
Correctly parse only changed markdown files on the second `watch` loop

bhagany:
Doing a full compare caused all the markdown to be re-parsed on the
second loop, because for some reason, the timestamp on the file
changed. Since we're only interested in the content of the file, just
comparing the file's hash to the previous version is sufficient, and
produces the correct behavior.


podviaznikov:
looks good and makes sense to me

PR #121, opened 2016-12-07
Refactor yaml

bhagany:
This is only an organizational change, intended to make the yaml parsing available to future input parsers.


podviaznikov (approved):

podviaznikov:
Makes sense to me

PR #122, opened 2016-12-07
Refactor and abstract rendering with `render-pre-wrap`

bhagany:
This is a subset of the changes in #109 that seemed to get the most positive response. It's almost purely a refactor intended to make rendering more pluggable. The only new feature is the addition of tracing (aka `:io.perun/trace` metadata) to `render` and `collection`.

### DIFF
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -558,14 +558,17 @@
   (let [options (merge +collection-defaults+
                        (dissoc *opts* :page)
                        (if-let [p (:page *opts*)]
-                         {:grouper #(-> {p {:entries %}})}
+                         {:grouper #(-> {p {:entries %
+                                            :group-meta (:meta *opts*)}})}
                          (if-let [gb (:groupby *opts*)]
                            {:grouper #(->> %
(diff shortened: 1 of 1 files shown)


podviaznikov:
Need to test this locally myself first. Give me a few days. 
The only braking change here is the removal of `:groupBy` option from the collection task. Everything else should work the same way, is it correct?

bhagany:
Take your time, I know I'm causing a lot of work :).  I actually left the `:groupby` option in this branch, I think it should stay unless we add `assortment`, and so everything should work the same as before.  The default `:groupby` is handled in the task instead of the `+collection-defaults+` map, in order to handle passing `:meta` to `render-pre-wrap`.  That deletion may have made it look like I removed `:groupby`.

podviaznikov:
@bhagany I merged some other PRs and they cause conflict here. Can you please resolve it when you have time? 

I tested all other PRs and they work for me quite nice. I really like `:io.perun/trace`

bhagany:
Okay, it's resolved.  Thanks!

PR #123, opened 2016-12-07
Use boot file metadata instead of maintaining a separate map

bhagany:
The most immediate benefit of this change is that `markdown` no longer needs to manually track files removed from the fileset between invocations. As a result, `markdown` is shorter and simpler, and the same will be true of any future input parsing tasks.

Some discussion is necessary on this one. There is a problem about how to deal with requiring `boot` in `io.perun.core`. The problem arises when a project doesn't depend on boot itself, and then executes a fn in a pod, and the fn's namespace requires `io.perun.core`. The pod can't find boot.core, and so, boom. Options:
 - Make projects depend on boot
 - Add boot to pod dependencies when pod namespaces require io.perun.core
 - Split the boot-using parts of io.perun.core into a separate namespace, and don't require that namespace from a pod
 - Something else?

I prefer the second option, but not strongly, and I'm not sure if there are performance considerations with adding more dependencies to pods.

### DIFF
--- a/src/io/perun/core.clj
+++ b/src/io/perun/core.clj
@@ -2,42 +2,8 @@
   "Utilies which can be used in base JVM and pods."
   (:require [clojure.java.io         :as io]
             [clojure.string          :as string]
-            [boot.core               :as boot]
             [boot.from.io.aviso.ansi :as ansi]
             [boot.util               :as u]))
-
-(def +meta-key+ :io.perun)
-
-(defn get-meta
-  "Return metadata on files. Files metadata is a list.
-   Internally it's stored as a map indexed by `:path`"
-  [fileset]
-  (keep +meta-key+ (vals (:tree fileset))))
-
-(defn key-meta [data]
-  (into {} (for [d data] [(:path d) d])))
-
-(defn set-meta
-  "Update `+meta-key+` metadata for files in `data` and return updated fileset"
-  [fileset data]
-  (boot/add-meta fileset (into {} (for [d data] [(:path d) {+meta-key+ d}]))))
-
-(defn merge-meta* [m1 m2]
-  (vals (merge-with merge (key-meta m1) (key-meta m2))))
-
-(defn merge-meta [fileset data]
-  (set-meta fileset (merge-meta* (get-meta fileset) data)))
-
-(def +global-meta-key+ :io.perun.global)
-
-(defn get-global-meta
-  "Return global metadata that is related to the whole project
-   and all files. Global metadata is a map"
-  [fileset]
-  (-> fileset meta +global-meta-key+))
-
-(defn set-global-meta [fileset data]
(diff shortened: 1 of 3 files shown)


src/io/perun/core.clj
```diff
@@ -11,15 +12,15 @@
   "Return metadata on files. Files metadata is a list.
    Internally it's stored as a map indexed by `:path`"
   [fileset]
-  (-> fileset meta +meta-key+ vals))
+  (keep +meta-key+ (vals (:tree fileset))))
```
podviaznikov: what is keep function?

src/io/perun/core.clj
```diff
@@ -11,15 +12,15 @@
...
 
 (defn key-meta [data]
   (into {} (for [d data] [(:path d) d])))
 
 (defn set-meta
-  "Update `+meta-key+` metadata for the fileset and return updates fileset"
+  "Update `+meta-key+` metadata for files in `data` and return updated fileset"
   [fileset data]
-  (vary-meta fileset assoc +meta-key+ (key-meta data)))
+  (boot/add-meta fileset (into {} (for [d data] [(:path d) {+meta-key+ d}]))))
```
podviaznikov: is this new boot function or it was there for a long time?

src/io/perun/core.clj
bhagany: `(keep :key coll)` is like `(filter identity #(map :key coll))` https://clojuredocs.org/clojure.core/keep

src/io/perun/core.clj
bhagany: I had to look it up, I am pretty new to boot.  Here is the commit that added it https://github.com/boot-clj/boot/commit/1ebcbb8dd78ee4fb88dad70e629630f9442a8b7a.  I think this means it has been in boot since 2.0?

bhagany: actually, on a second look, I think the function was just moved in this commit.  Here is the real commit: https://github.com/boot-clj/boot/commit/038d8dde09704e3d22b5cdfdd8be984825f04bf9

src/io/perun/core.clj
bhagany: Perhaps I should explain - `get-meta` returns data in the same shape as before - a seq of metadata maps.  This PR doesn't have any API changes, aside from not needing to deal with removed markdown files.

Deraen:
> Split the boot-using parts of io.perun.core into a separate namespace, and don't require that namespace from a pod

Is probably the correct solution. It is generally best to presume that `boot.core` is not available in pods.

bhagany:
Okay @Deraen, sounds good to me.  I've committed that change.

podviaznikov (approved):

podviaznikov:
Looks good to me. Merging.

PR #130, opened 2017-01-01
Fix #129 - Make `draft` do the correct thing

### LINKED ISSUE #129: draft task broken in master
Probably due to recent refactorings, the `draft` task is currently broken: it doesn't actually cause the drafts not to be rendered (or shown in collections).

Here's a version that works for me:

```clojure
(deftask remove-draft
  []
  (with-pre-wrap fileset
    (let [files (filter identity (pm/get-meta fileset))
          updated-files (->> files
                             (map #(cond-> % (:draft %) (dissoc :content :include-atom :include-rss))))]
      (pm/set-meta fileset updated-files))))
```

bhagany:
This is a pretty simple change, but I'd like to give people a chance to take a look, because it's a departure from how it used to work.  Now that we're storing metadata on the boot tmpfiles themselves, removing the file also removes its metadata, and that's what I chose to do here.  It seems cleaner to me to do it that way, rather than just deleting some metadata.

Testing this change also revealed a bug in the way markdown was merging old and new metadata.  It was a small change, so I fixed that as well.

### DIFF
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -143,8 +143,9 @@
             updated-files (trace :io.perun/markdown
                                  (pod/with-call-in @pod
                                    (io.perun.markdown/parse-markdown ~md-files ~options)))
-            final-metadata (pm/merge-meta* (pm/get-meta @prev-fs) updated-files)
-            new-fs (pm/set-meta fileset final-metadata)]
+            new-fs (-> fileset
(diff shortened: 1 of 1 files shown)


podviaznikov:
That looks great. I recently hit problem with draft task myself and was thinking of removing it since it wasn't working correctly. I'm glad there is a proper fix.

PR #131, opened 2017-01-04
Merge filedata

bhagany:
If you're wondering if I have a series of branches just waiting to be pull requests... well, yes, I do.  I even wiki'd it to keep track, if you're interested: https://github.com/bhagany/perun/wiki

Anyway, this is a simple change.  I was always surprised that `add-filedata` overwrites all of the existing metadata for the files it touches.  This PR merges the metadata instead, so that you don't have to worry about having tasks touch the same file and deleting stuff you expected to be there.  I can't think of any use cases that this would break, but if you have some, I'd be interested to hear.

### DIFF
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -56,7 +56,7 @@
 (defn add-filedata [tmp-files]
   (pod/with-call-in @filedata-pod
     (io.perun.filedata/filedatas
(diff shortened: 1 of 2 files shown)


bhagany:
By the way, I can squash these if you'd like, just let me know.

podviaznikov (approved):
Looks good to me

podviaznikov:
Great write-up here https://github.com/bhagany/perun/wiki/Content-pipeline.

Issue #77, opened 2015-12-02
Collection task should work even if no :content files are found


podviaznikov:
E.x. https://github.com/hashobject/perun.io/blob/master/build.boot#L32 this should work without creation of fake `index.md`

martinklepsch:
I'm thinking the same should be true for the render task. If the user only wants to render files that contain a `:content` key they may use the `filterer` option. Hardcoded filtering like this makes it harder to use the render task to render files other than markdown. (i.e. `edn`)

podviaznikov:
Here is the problem when there are not metadata files at all and I want to render one page. And I would want probably to use collection for that.

E.x. I wanted to create static site for perun and collection didn't work because I didn't have any initial files that should be parsed and added to the meta

bhagany:
`collection` now works without any input files present, as of https://github.com/hashobject/perun/commit/341ed5c54552608144cbb9b364253639874d8b07

PR #132, opened 2017-01-08
Content pipeline

bhagany:
Credit to @MartyGentillon for the ideas that led to this PR (see the conversation on https://github.com/hashobject/perun/pull/109). This is the big change most of my past PR's have been leading up to - it meets the following goals:

- support existing usage patterns
- allow more possible compositions of perun tasks
- allow 3rd party tasks to be interposed with perun tasks more easily
- resolve weirdness about the `:content` key
- abstract content parsing similarly to the rendering abstraction (aka `render-pre-wrap`)

The central conceptual change here is completing the transition from a more metadata-based approach to a more file-based approach. The first major step here was to store file metadata on boot's TmpFiles, rather than in fileset metadata. This allows us to move or delete files without having to maintain their metadata separately. The second (and last) step, is to do away with the `:content` metadata key in favor of writing and manipulating TmpFiles, and using the content of the files, rather than maintaining that information in metadata. This allows any task to manipulate perun's final output, without that task having to be actually aware of perun and its metadata conventions.

For example, I know some here are fans of Tachyons (http://tachyons.io/). With this change, it would be possible to write a _general_ `tachyons` task that manipulates file content and can be used with or without perun. To explain with code,

```clojure
(deftask build
  (comp (p/markdown)
        (tachyons) ;; <-- unaware of perun
        (p/render :renderer 'your.ns/render)))
```

Before this change, the `tachyons` task would need to know about and manipulate perun's `:content` key in order to get what you wanted in `perun/render`, but now it can just change the file itself and ignore perun.

In addition to being more composable with 3rd party tasks, the change from metadata content to file content also allows perun's task to be composed with each other in ways that weren't possible before. Here's a real-life example from the `build.boot` for my blog\*:

```clojure
(deftask urls
  "Helper task for encapsulating all url generation stuff"
  [_ filterer FILTERER code "predicate to use for selecting entries (default: `:has-content`)"]
  (let [filterer (or filterer :has-content)]
    (comp (p/slug :slug-fn slugify-filename :filterer filterer)
          (p/permalink :permalink-fn permalinkify :filterer filterer)
          (p/canonical-url :filterer filterer))))

(deftask build
  "Build nicerthantriton.com"
  [t tier TIER kw "The tier we're building for"]
  (comp (p/global-metadata)
        (set-tier :val tier)
        (p/markdown :options {:extensions {:smarts true}})  ;; writes an .html file per .md file
        (urls)
        (p/collection :renderer 'nicerthantriton.core/index ;; produces an index page based on the output from `markdown`
                      :filterer post?
                      :meta {:derived true})
        (p/assortment :renderer 'nicerthantriton.core/topic ;; not in master yet, you can ignore it
                      :filterer post?
                      :grouper tagify
                      :meta {:derived true})
        (urls :filterer :derived)
        (recent-posts)                                      ;; These two set some global metadata based on .html files
        (topics)                                            ;; that exist at this point in the pipeline - you can ignore
        (p/render :renderer 'nicerthantriton.core/post      ;; Wraps post content with things like tags and publish date
                  :filterer post?)
        (p/render :renderer 'nicerthantriton.core/page)     ;; Wraps _all_ .html files with the things that are common to every page
        (p/atom-feed :filterer post?)))
```

Some interesting things about this:
- `collection` can be called before `render`, and the result can be treated the same as files generated by `markdown`
- multiple `render` tasks can be used to progressively build up pages
- the original usage pattern is still supported - that is, you can still use a single `render` step, and `collection` could render a finished page in one pass, if that works best for you.

I'm very interested in any feedback you have, and in particular, if there are any use cases that are currently supported in master, but unsupported here.

Lastly, I would like you all to know that every time I think about this branch, my brain sings "It's the content pipeline" to the tune of "It's the final countdown". You're welcome.

\* There's no content on the blog yet. Working on it.

### DIFF
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -136,34 +136,26 @@
                                boot/user-files
                                (boot/by-ext file-exts)
                                add-filedata)
-            changed-meta (->> (if pod
-                                (pod/with-call-in @pod ~(parse-form changed-files))
-                                (eval (parse-form changed-files)))
-                              (map #(let [{:keys [include-rss include-atom]} %]
-                                      (-> %
-                                          (assoc :include-rss* include-rss
-                                                 :include-atom* include-atom)
-                                          (dissoc :include-rss :include-atom))))
-                              (trace tracer))
+            changed-meta (trace tracer
+                                (if pod
+                                  (pod/with-call-in @pod ~(parse-form changed-files))
+                                  (eval (parse-form changed-files))))
             input-fs (-> (if @prev-fs
                            (pm/set-meta fileset (meta-by-ext @prev-fs file-exts))
                            fileset)
                          (pm/set-meta changed-meta))
             input-meta (meta-by-ext input-fs file-exts)
             output-meta (doall
-                         (for [{:keys [path parsed filename
-                                       include-rss* include-atom*] :as entry*} input-meta]
+                         (for [{:keys [path parsed filename] :as entry*} input-meta]
                            (let [page-filepath (string/replace path #"(?i).[a-z]+$" ".html")
                                  entry (-> entry*
                                            (assoc :has-content true
                                                   :original-path path
                                                   :path page-filepath
                                                   :filename (string/replace filename
-                                                                            #"(?i).[a-z]+$" ".html")
-                                                  :include-rss include-rss*
-                                                  :include-atom include-atom*)
+                                                                            #"(?i).[a-z]+$" ".html"))
                                            (dissoc :parsed :extension :file-type :full-path
-                                                   :mime-type :original :include-rss* :include-atom*))]
+                                                   :mime-type :original))]
                              (perun/create-file tmp page-filepath parsed)
                              entry)))
             new-fs (-> input-fs
@@ -406,22 +398,28 @@
 (def ^:private +rss-defaults+
   {:filename "feed.rss"
    :filterer :include-rss
+   :extensions [".html"]
    :out-dir "public"})
 
 (deftask rss
   "Generate RSS feed"
-  [f filename    FILENAME    str  "generated RSS feed filename"
-   _ filterer    FILTER      code "predicate to use for selecting entries (default: `:include-rss`)"
-   o out-dir     OUTDIR      str  "the output directory"
-   t site-title  TITLE       str  "feed title"
-   p description DESCRIPTION str  "feed description"
-   l base-url    LINK        str  "feed link"]
+  [f filename    FILENAME    str   "generated RSS feed filename"
+   _ filterer    FILTER      code  "predicate to use for selecting entries (default: `:include-rss`)"
+   e extensions  EXTENSIONS  [str] "extensions of files to include in the feed"
+   o out-dir     OUTDIR      str   "the output directory"
+   t site-title  TITLE       str   "feed title"
+   p description DESCRIPTION str   "feed description"
+   l base-url    LINK        str   "feed link"]
   (let [pod (create-pod rss-deps)
         tmp (boot/tmp-dir!)]
     (boot/with-pre-wrap fileset
       (let [global-meta   (pm/get-global-meta fileset)
             options       (merge +rss-defaults+ global-meta *opts*)
-            files         (filter (:filterer options) (pm/get-meta fileset))]
+            files         (->> fileset
+                               boot/user-files
+                               (boot/by-ext (:extensions options))
+                               (keep pm/+meta-key+)
+                               (filter (:filterer options)))]
         (perun/assert-base-url (:base-url options))
         (pod/with-call-in @pod
           (io.perun.rss/generate-rss ~(.getPath tmp) ~files ~(dissoc options :filterer)))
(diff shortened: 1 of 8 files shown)


bhagany:
I should also add:

- this has been in progress awhile, I can squash if you prefer :)
- it fixes #77 (Collection task should work even if no :content files are found)
- makes #105 (pre-parse function to markdown task) easy to implement yourself as a standalone task

podviaznikov:
Code looks good to me. Will need to test it locally first. Thanks for the detailed description

bhagany:
Correction: #105 still isn't really possible with this PR.  However, I did just complete a branch for a future PR that does fulfill that goal.  All in good time :).  See https://github.com/bhagany/perun/wiki for details if you're curious.

src/io/perun.clj
```diff
@@ -116,6 +116,54 @@
...
+                                  (pod/with-call-in @pod ~(parse-form changed-files))
+                                  (eval (parse-form changed-files))))
+            input-fs (-> (if @prev-fs
+                           (pm/set-meta fileset (meta-by-ext @prev-fs file-exts))
+                           fileset)
+                         (pm/set-meta changed-meta))
+            input-meta (meta-by-ext input-fs file-exts)
+            output-meta (doall
+                         (for [{:keys [path parsed filename] :as entry*} input-meta]
+                           (let [page-filepath (string/replace path #"(?i).[a-z]+$" ".html")
```
MartyGentillon: Extensions can contain nearly any character including '-' and '_'.  Perhaps this should be constructed by the file-exts parameter?

src/io/perun.clj
```diff
@@ -116,6 +116,54 @@
...
+                                 entry (-> entry*
+                                           (assoc :has-content true
+                                                  :original-path path
+                                                  :path page-filepath
+                                                  :filename (string/replace filename
+                                                                            #"(?i).[a-z]+$" ".html"))
```
MartyGentillon: See above.

src/io/perun.clj
```diff
@@ -116,6 +116,54 @@
...
+                                           (dissoc :parsed :extension :file-type :full-path
+                                                   :mime-type :original))]
```
MartyGentillon: It might be better to update the `:mime-type`, `:extension`, and `:file-type` keys rather than removing them.  They seems like something that it would be useful to filter on.

src/io/perun.clj
```diff
@@ -129,26 +177,17 @@
   "Parse markdown files
 
    This task will look for files ending with `md` or `markdown`
-   and add a `:content` key to their metadata containing the
+   and add a `:parsed` key to their metadata containing the
    HTML resulting from processing markdown file's content"
```
MartyGentillon: It seems that this also writes out HTML files to the location related, is that correct?

src/io/perun.clj
```diff
@@ -455,15 +524,28 @@
...
+  (let [tmp (boot/tmp-dir!)]
     (boot/with-pre-wrap fileset
-      (let [new-metadata (-> fileset
-                             (render-paths-fn options)
-                             (render-to-paths (:renderer options) tmp tracer))]
+      (let [render-paths (render-paths-fn fileset options)
+            new-metadata (render-to-paths render-paths (:renderer options) tmp tracer)
+            rm-files (keep #(boot/tmp-get fileset (-> % :entry :path)) (vals render-paths))]
         (-> fileset
+            (boot/rm rm-files)
```
MartyGentillon, line 539: It might makes sense to log the names of removed files in debug mode.

src/io/perun.clj
```diff
@@ -481,28 +563,22 @@
...
-                                                  (or (some-> (:permalink entry)
-                                                              (string/replace #"/$" "/index.html")
-                                                              perun/url-to-path)
-                                                      (string/replace path #"(?i).[a-z]+$" ".html")))]
+                           (fn [result {:keys [path permalink] :as entry}]
+                             (let [content (slurp (boot/tmp-file (boot/tmp-get fileset path)))
+                                   new-path (make-path (:out-dir options) permalink path)
+                                   meta-entry (merge meta entry)
+                                   content-entry (assoc meta-entry :content content)]
                                (perun/report-debug "render" "rendered page for path" path)
```
MartyGentillon: This logging message is potentially confusing, as the rendering isn't actually occurring here.  When someone's render task is crashing they might think that it had worked given that this printed out for all files.  Perhaps it should be moved to render-to-paths.

bhagany:
Excellent feedback, @MartyGentillon.  I believe I've addressed all of your concerns.

MartyGentillon (approved):
Looks good to me.

bhagany:
I merged #134 here, because I assumed it would get merged before this one.  This way, there'll be no conflicts here when that happens.  If there's a problem with #134 that I can't resolve quickly, I'll undo this merge.

bhagany:
(also, yes, the tests pass here)

podviaznikov (approved):
Looks great to me

podviaznikov:
@MartyGentillon thank you for your help with discussion and reviews

PR #134, opened 2017-01-14
A basic test suite for tasks

bhagany:
This is the sort of thing I'd probably merge myself, but there are several places where I used `boot/ls` rather than `boot/user-files` in the code proper, in order to make testing easier.  I'm pretty sure this won't break anyone's use cases, but since it might, I thought it best to let others review it.

But hey!  Tests!  Run them with `boot runtests`.


podviaznikov (approved):
Nice, tests!

Issue #99, opened 2016-09-08
`permalink` depends on `slug`


pesterhazy:
By default, the `permalink` task only works if the `slug` ask was run previously. This caught me off-guard.

Possible solutions:
- call slug if metadata key not already present?
- throw an exception at runtime if slog is not present
- warn at runtime if slug is not present
- add documentation

martinklepsch:
There's the `SPEC.md` file which notes this dependency but we should definitely become smarter about errors in this context.  Slightly elm-inspired here's an idea for an error that could be shown when the slug is missing. 

![screenshot 2016-09-08 17 55 23](https://cloud.githubusercontent.com/assets/97496/18354326/cc6b564e-75ed-11e6-8919-9c62c685a29f.png)

Because slug-fn and similar parts of perun are swappable we can't do something fully generic at the task level I believe but maybe we can use `clojure.spec` or similar on default fn's to give this kind of feedback.

podviaznikov:
I like this type of feedback. Looks super beginner friendly

Deraen:
And we could probably use puget to colorize the output + shorten long fields (string over 80 chars or such), so we don't manually need to dissoc long fields (content).

Deraen:
And this is once again a place we would need to throw a exception to stop Boot pipeline, but showing the stacktrace wouldn't make much sense. But this should be fixed in the next Boot version: https://github.com/boot-clj/boot/issues/486

martinklepsch:
This is what I used to omit long values :D

``` clojure
(defn omit-long-vals [m]
  (reduce-kv (fn [m k v]
               (if (and (string? v) (< 150 (count v)))
                 (assoc m k ::omitted)
                 m))
             m
             m))
```

bhagany:
Fixed as of https://github.com/hashobject/perun/commit/4638cd5baa1a2ff8d7d3c0e8293cb2264d60b3e1.

`:slug`, `:permalink`, and `:canonical-url`\* are now based on a file's path in the fileset, and automatically set for you when metadata is accessed.  In many cases, this means you don't need to run the `slug` and `permalink` tasks at all, but you still can, if you want to move the files to new locations, based on a slug or permalink.  The `canonical-url` task is deprecated, because there is no advantage to moving a file given a new canonical url, over a new permalink.

\* note, you still must have the `:base-url` global metadata set in order for `:canonical-url` to be set for you.

PR #138, opened 2017-01-16
Fileset Invictus

bhagany:
This is the last big PR you'll have to deal with from me for a while - from here my plans are mostly to close tickets and write docs.  I'd like to say thank you to everyone who has looked at these PR's and offered their opinions - I know that I've been causing a lot of work for everyone, and I really appreciate the time you've put in. The support and help I've received have been great, and I hope you feel as good about where Perun is going as I do.

If `content-pipeline` was the "tick", then this PR is the "tock".  Whereas the focus of `content-pipeline` was to favor manipulating the contents of files over tracking metadata, here the focus in on file extension and location over metadata. Also, I split these changes into two PR's because while `content-pipeline` is backwards compatible, this PR is not (don't worry, the api changes are minor). The goals here are:

- Increase the range of things 3rd party tasks can do without Perun's knowledge, ie. move, rename or create files
- Resolve coupling between `slug`, `permalink`, and `canonical-url` (#99)
- Stop relying on too-general metadata (ie. `:has-content`) for filtering

My jumping off point for these changes was my discomfort with the path-related tasks, primarily `slug`, `permalink` and `canonical-url`. As #99 suggests, the coupling between them is not ideal. The thing that first got to me was that it's possible to end up with inconsistencies between these values. But more than that, the values that these tasks set are functions of values that Perun already knows\*. It occurred to me that if Perun were to use a file's location in the fileset as the source of truth for these derived values, then the coupling problem melts away. In fact, in many cases, the need to use these tasks at all goes away.

Rather than setting path-related values in metadata the first time Perun visits a file, this PR instead automatically sets them when metadata is accessed. This guarantees that the values are consistent with the location of the files they were retrieved from. In addition, the `base` task and the `add-filedata` helper function are now obsolete, so `base` is a pass-thru and outputs a deprecation warning. One consequence of removing `add-filedata` is that Perun no longer automatically sets the `:mime-type` and `:file-type` metadata keys. These weren't used by Perun itself, and in case anybody relied on them, I made a new `mime-type` task for setting these keys.

For flexibility, I added the `:doc-root` key to the global metadata. Where `:out-dir` is the part you want in the front of `:path`, `:doc-root` is the part you _don't_ want in the `:permalink` and `:canonical-url`.  Its value will almost always be the same as `:out-dir`, but in cases where you'd like to output to subdirectories of your document root, it's helpful to distinguish between the two.  `:doc-root` defaults to `"public"`, just like `:out-dir`.

In the interest of backward compatibility, I kept the semantics of `slug` and `permalink` the same. The difference is that now, these tasks move files rather than setting metadata. Also, because all files implicitly have a `:slug` and a `:permalink` by virtue of having a `:path`, running these tasks is no longer a prerequisite for other tasks. The only reason to use them is to move files, and any other task that moves files (like `sift`) will have exactly the same effect. I considered modifying the `canonical-url` task to move files as well, but moving a file given a new `:canonical-url` doesn't have any advantages over doing so with a new `:permalink`, so I deprecated `canonical-url` in the same manner as `base`. All files will still implicitly have `:canonical-url` if a `:base-url` is set in global metadata.

To further enhance Perun's composability with 3rd parties, I have added an `extensions` parameter to every task I could. This allowed me to remove the `:has-content` key, defaulting to `identity` for all filterers instead. This means that, by default, any 3rd party task can create files that go on to participate in Perun's tasks, in the same manner as files that Perun created. For example, let's say you already had a task that produced an HTML file, and want to continue to use this task with Perun. You can do something like this, and your custom HTML will be picked up by Perun by default:

```clojure
(deftask build
  (comp (p/markdown)
        (your-custom-html-task) ;; <-- produces a file with no Perun metadata attached
        (p/render :renderer 'your-ns/wrap-content) ;; <-- wraps the output of `markdown` AND `your-custom-html-task`
        ))
```

In addition to these changes, I did some general tightening up of the code. If you have questions about some choices I made, feel free to ask. I am happy to explain my reasoning, and to change when it's wrong or undesired.

### Summary of backwards incompatible changes
- if you relied on `:mime-type` or `:file-type`, the new `mime-type` task should be added to your pipeline
- if your `:out-dir` is something other than `"public"`, you'll probably want to set `:doc-root` in global metadata
- if you've written filters that are supposed to select non-HTML files, you'll have to add an `extensions` parameter that includes those files
- the `permalink-fn` and `slug-fn` functions now have the same signature, and they take two parameters - `global-meta`, and the metadata for the entry.

\* `:slug` is a function of `:path`, `:permalink` is a function of `:path` and `:out-dir`\*\*, and `:canonical-url` is a function of `:base-url` and `:permalink`.

\*\* `:permalink` _used_ to be a function of `:out-dir`, but this PR makes it a function of the new `:doc-root` global metadata.


src/io/perun.clj
```diff
@@ -27,47 +29,88 @@
...
+  [fileset options]
+  (filter (comp (:filterer options) (partial pm/meta-from-file fileset))
+          (tmp-by-ext fileset (:extensions options))))
+
+(defn filter-meta-by-ext
+  "Returns perun metadata from `fileset`. `options` selects files
+  that end with values in the `:extensions` key, filtered by the
+  `:filterer` predicate. If `:extensions` is empty, returns
+  metadata for all files."
+  [fileset options]
```
MartyGentillon: Given that you are only using the :filterer and :extensions keys from options, a better signature might be `[fileset {:keys [filterer extensions]}]` or perhaps `[fileset {:keys [filterer extensions] :as options}]`.  (same goes for above.)  

Also, `filter-fn` might (or might not) be a better name for filterer.

src/io/perun.clj
```diff
@@ -105,74 +145,103 @@
...
-                                  (eval (parse-form changed-files))))
-            input-fs (-> (if @prev-fs
-                           (pm/set-meta fileset (meta-by-ext @prev-fs file-exts))
-                           fileset)
+      (let [global-meta (pm/get-global-meta fileset)
+            prev-fs (:fs @prev)
+            parse-form (parse-form-fn (meta-by-ext (boot/fileset-diff prev-fs fileset :hash) extensions))
+            ;; Change `diff` to boot/fileset-diff when
+            ;; https://github.com/boot-clj/boot/pull/566 is merged
+            changed-from-prev-meta (meta-by-ext (diff prev-fs fileset pm/+meta-key+) extensions)
```
MartyGentillon: So, you have computed the diff twice here, and the code doesn't make it clear why.  Part of this has to do with naming: `changed-from-prev-meta` isn't terribly enlightening.  A second part has to do with data flow: it looks like the Perun metadata is being passed around the parsing function, making it unavailable for parsing control.  Either way, the code feels somewhat confusing.

src/io/perun.clj
```diff
@@ -105,74 +145,103 @@
...
-  and writes html files for subsequent tasks to process, if desired. Pass
+  files with extensions in `extensions`, adds `tracer` to `:io.perun/trace`
+  and writes files for subsequent tasks to process, if desired.
+  `output-extension` can be used to indicate the type of file the task produces,
+  or pass `nil` to overwrite with a file of the same name and extension. Pass
   `pod` if one is needed for parsing"
-  [parse-form file-exts tracer & [pod]]
-  (let [tmp     (boot/tmp-dir!)
-        prev-fs (atom nil)]
+  [parse-form-fn extensions output-extension tracer options & [pod]]
```
MartyGentillon: This has enough arguments that pass-by-name is probably a better approach.

src/io/perun.clj
```diff
@@ -216,189 +288,200 @@
...
-                             (map #(assoc % :slug (-> % :filename slug-fn)))
-                             (trace :io.perun/slug))]
-      (perun/report-debug "slug" "generated slugs" (map :slug updated-files))
-      (perun/report-info "slug" "added slugs to %s files" (count updated-files))
-      (pm/set-meta fileset updated-files))))
+  [s slug-fn    SLUGFN     code  "function to build slug from file metadata"
+   _ filterer   FILTER     code  "predicate to use for selecting entries (default: `identity`)"
+   e extensions EXTENSIONS [str] "extensions of files to include"]
+  (let [options (merge +slug-defaults+ *opts*)
+        slug-fn (:slug-fn options)
```
MartyGentillon: consider [{:keys [slug-fn] :as options}] (merge +slug-defaults+ *opts*)

src/io/perun.clj
```diff
@@ -216,189 +288,200 @@
...
-      (perun/report-info "permalink" "added permalinks to %s files" (count updated-files))
-      (pm/merge-meta fileset updated-files))))
-
-(def ^:private +canonical-url-defaults+
-  {:filterer :has-content})
+  [p permalink-fn PERMALINKFN code  "function to build permalink from TmpFile metadata"
+   _ filterer     FILTER      code  "predicate to use for selecting entries (default: `identity`)"
+   e extensions   EXTENSIONS  [str] "extensions of files to include"]
+  (let [options (merge +permalink-defaults+ *opts*)
+        permalink-fn (:permalink-fn options)
```
MartyGentillon: consider destructuring.

src/io/perun.clj
```diff
@@ -594,23 +661,19 @@
   on the provided `fileset` and `options`."
   [task-name fileset options]
   (let [global-meta (pm/get-global-meta fileset)
-        grouper (:grouper options)
-        paths (->> fileset
-                   pm/get-meta
-                   (filter (:filterer options))
-                   grouper)]
+        {:keys [grouper sortby comparator out-dir]} options
```
MartyGentillon: This should be in the method signature like so `[task-name fileset {:keys [grouper sortby comparator out-dir] :as options}]'

src/io/perun/meta.clj
```diff
@@ -1,41 +1,75 @@
...
+            [clojure.java.io :as io]
+            [clojure.string :as string]
+            [io.perun.core :as perun]))
 
 (def +meta-key+ :io.perun)
 
+(def +global-meta-key+ :io.perun.global)
+
+(def +global-meta-defaults+
+  {:doc-root "public"})
```
MartyGentillon, line 13: Maybe consider a dynamic var?

src/io/perun.clj
bhagany: Changing `:filterer` would be backwards-incompatible, as these options are passed through from the public-facing API.  I don't really have an opinion on the name, but I was trying to keep API changes to a minimum.  I don't mind changing it if a few more people chime in on it, however.

Noted on the fn signature.

MartyGentillon: fair enough.  (you could also use the destructuring to rename the variable without changing the key.)

src/io/perun.clj
bhagany: I'll go with a map - I don't much like the `(a-fn :key1 "foo" :key2 "bar")` style.  Feel free to voice any objections to that, though.

MartyGentillon: Either one fixes the problem where when I read the call I wonder "is out-ext the third or fourth argument?"

src/io/perun.clj
bhagany: This is confusing, and it's taken me many iterations to get it functionally correct, which has harmed the readability.  To answer your immediate question, the code actually calculates two different diffs.  The first diff gives back files whose content has changed, and the second diff gives back files whose perun metadata has changed.  `:hash` vs `pm/+meta-key+` is what makes the calls different.  `changed-from-prev-meta` is supposed to indicate "files whose metadata has changed since the previous fileset we saw".

Anyway, I'll give it another readability pass tomorrow.  Suggestions on easier reading are welcome.

src/io/perun/meta.clj
bhagany: Could you explain your reasoning on the advantages of this?  It's not immediately clear to me.

src/io/perun.clj
MartyGentillon: That makes sense, though makes me wonder if the files where the Perun metadata has changed might need to be reparsed.  As for making it slightly clearer perhaps name the diffs as follows: `content-changed-files` and `meta-changed-files` (or something similar).

bhagany:
Thanks for another thorough review, @MartyGentillon.  Do you have any thoughts on the general approach?  I am very happy with it, but I wonder if there are things I'm not seeing.

src/io/perun/meta.clj
MartyGentillon: First, I am not at all sure that this is a good idea (there are probably other ways to change the default values, such as putting them in the global meta file).  

But the basic question is "how does someone change this?" when reading the code, a dynamic var would provide a clear answer.

MartyGentillon:
Nope, the basic strategy is pretty much exactly where I was hoping this whole thing was going to move.  If you hadn't done it, I probably would have moved in exactly the same direction.

src/io/perun/meta.clj
bhagany: Okay, I'll give it some thought, then.  My (not really carefully considered) reasoning for putting `:doc-root` in global metadata was that it's a counterpart to `:base-url`, which is also global metadata.  I think this probably holds up, now that I'm thinking about it more. I'd advocate for doing the same thing for `:base-url` and `:doc-root`, whatever that may be.

src/io/perun.clj
bhagany: > That makes sense, though makes me wonder if the files where the Perun metadata has changed might need to be reparsed.

I've had some thoughts in this direction, actually.  I'll try to lay them out briefly.

1. `markdown`, and other content parsing tasks that are in progress or have been discussed, don't use Perun metadata in any meaningful way.
1. In a future PR (here: https://github.com/bhagany/perun/tree/markdown-asunder), I've split the `markdown` task into yaml and markdown-only subtasks, as you argued for here: https://github.com/hashobject/perun/pull/106
1. If that PR is merged, it's then possible to intelligently skip markdown parsing when only the yaml metadata has changed in a file.  I liked that, so I laid the groundwork here, which means diffing content and metadata separately.
1. However, I've also had thoughts about merging `content-pre-wrap` and `render-pre-wrap`, since they are so conceptually similar. If that happens, then skipping processing when metadata changes will no longer be feasible.
1. Adding a content parser that uses Perun metadata would also make the two diffs approach infeasible.

I guess I'm going with a do-the-most-efficient-thing-until-we-can't approach?

MartyGentillon: Seams reasonable.  Also, when/if the until-we-can't part happens, a parameter could control that part of the behavior.

bhagany:
Github is hiding @MartyGentillon's review after I pushed some new commits, so I'm moving a few items for discussion to the main thread.

- I've tried to make `content-pre-wrap` a bit more readable, and it probably needs some more work in that direction.  I'd appreciate suggestions.
- We briefly discussed whether `content-pre-wrap` should trigger re-renders when Perun metadata changes, and @MartyGentillon suggested controlling this behavior with a parameter. This seemed easy to do now, so I added a `uses-meta` flag to `content-pre-wrap`.
- The possibility of making `doc-root` a dynamic var is up for discussion.  I'm open to the idea, but I think that if we do that, `base-url` should be a dynamic var too, as they are conceptually similar.

podviaznikov (approved):
This looks great. Thank you so much for detailed explanation. 
As a side note I'm very happy that we don't need base task anymore. I didn't feel good about it but also didn't know how to get rid of it.

podviaznikov:
Also thank you so much for trying in most cases to keep in backward compatible. I think it is also amazing that they did so many internal changes with very small changes of API. I'm even more positive about perun external model now.

bhagany:
> I'm even more positive about perun external model now.

I think I said this in Slack when I first started contributing - Perun's model has a lot of promise.  From the start, I liked its focus on simple tasks that do few things (and usually just one thing, which is perfect), and the way it leveraged Boot's task model so effectively.  You're doing a great job, @podviaznikov, and ditto every one else who has contributed.  Thanks for some great software, and for making me feel welcome here.

test/io/perun_test.clj
```diff
@@ -30,15 +32,15 @@
...
 (deftask value-test
   [p path     PATH    str "path of the file to test"
    v value-fn VALUEFN edn "the value to test (optional)"
    m msg      MSG     str "message shown on failure"]
   (boot/with-pass-thru fileset
     (let [file (boot/tmp-get fileset path)]
-      (is (and (not (nil? file)) (value-fn file)) msg))))
+      (is (and (not (nil? file)) (value-fn fileset file)) msg))))
 
 (deftask content-test
```
MartyGentillon: These aren't tests so much as checks.

test/io/perun_test.clj
```diff
@@ -77,25 +79,25 @@
...
-        (value-test :path "public/test-image_200.jpg" :value-fn #(meta= % :height 200))
-        (value-test :path "public/test-image_200.jpg" :value-fn #(meta-contains? % :io.perun/trace :io.perun/images-resize))))
+        (value-test :path "public/test-image_100.jpg" :value-fn #(meta= %1 %2 :width 100))
+        (value-test :path "public/test-image_100.jpg" :value-fn #(meta= %1 %2 :height 100))
+        (value-test :path "public/test-image_100.jpg" :value-fn #(meta-contains? %1 %2 :io.perun/trace :io.perun/images-resize))
+        (value-test :path "public/test-image_200.jpg" :value-fn #(meta= %1 %2 :width 200))
+        (value-test :path "public/test-image_200.jpg" :value-fn #(meta= %1 %2 :height 200))
+        (value-test :path "public/test-image_200.jpg" :value-fn #(meta-contains? %1 %2 :io.perun/trace :io.perun/images-resize))))
 
 (deftask add-txt-file
```
MartyGentillon, line 132: It looks like a lot of these would be generally useful for testing.  We might consider moving them to a testing library so that other related libraries can use them.

MartyGentillon (approved):
A couple of thoughts about the tests for later.

src/io/perun.clj
```diff
@@ -177,41 +177,49 @@
...
+(defn diff-filesets
+  [before after uses-meta]
+  (if uses-meta
+    ;; Change `fileset-diff` to `boot/fileset-diff` when
+    ;; https://github.com/boot-clj/boot/pull/566 is merged
+    (let [diff (fileset-diff before after :hash pm/+meta-key+)]
+      {:content-diff diff
+       :meta-diff diff})
+    {:content-diff (fileset-diff before after :hash)
+     :meta-diff (fileset-diff before after pm/+meta-key+)}))
```
MartyGentillon: This fixes the main readability issue, but I am still not certain about the names. `content-diff` and `meta-diff` read as the diff of the content and meta respectively, whereas these are files that have changed content and meta respectively.

MartyGentillon: This fixes the main readability issue, but I am still not certain about the names. content-diff and meta-diff read as the diff of the content and meta respectively, whereas these are files that have changed content and meta respectively.

MartyGentillon (approved):

test/io/perun_test.clj
bhagany: I don't understand the distinction, could you elaborate?

src/io/perun.clj
bhagany: > content-diff and meta-diff read as the diff of the content and meta respectively

heh, this actually is how I think of them, but I see your point.  Another stab at it coming shortly.

test/io/perun_test.clj
MartyGentillon: A test exercises some functionality and then checks that it's behavior is correct.  A check or assert only makes sure that some invariant is true.

MartyGentillon (approved):

test/io/perun_test.clj
bhagany: Okay, makes sense to me, I'll change those names.

MartyGentillon (approved):

PR #140, opened 2017-01-22
Fix #27 - Tags

### LINKED ISSUE #27: New plugin: tags
I'm thinking about creating tags plugin like this (https://github.com/totocaster/metalsmith-tags).
@martinklepsch, @Deraen do you need tag pages for your blogs, or it's not important right now?

bhagany:
Provides a `tags` task that creates an html file for each tag found in the `:tags` metadata of existing html files.  The render function provided to the task will be called for each file to be generated a data structure in the same shape as the one provided to the `collection` task, a map with `:entry`, `:entries` and `:meta` keys.

The `assortment` task (first discussed in https://github.com/hashobject/perun/pull/109) comes along for the ride here, because `tags` is a subset of `assortment`.

This will probably conflict with other pull requests that I'm submitting at the same time, but they'll be trivial for me to fix as they get merged.


bhagany:
I should have mentioned - there is one backwards-incompatible change to `collection`.  The `groupby` option has been removed, because this is covered more natrually by `assortment`.  It can be added back in relatively easily if you prefer.

podviaznikov (approved):
Same comment as before.
---
Code looks good to me.
Only one ask would be to add use of assortment and paginate tasks into the example project https://github.com/hashobject/perun/tree/master/examples/blog

bhagany:
The last commit merged #141 here, for conflict avoidance.

podviaznikov (approved):

PR #141, opened 2017-01-22
Fix #63 - Paginate

### LINKED ISSUE #63: Pagination task
@martinklepsch I think we can extract your pagination task(https://github.com/martinklepsch/martinklepsch.org/blob/master/build.boot) and put it into perun. What do you think?

bhagany:
Provides a `paginate` task that sorts existing entries, partitions them by `page-size`, and renders an html file for each of the resulting partitions.  The render function provided to the task will be called for each file to be generated a data structure in the same shape as the one provided to the `collection` task, a map with `:entry`, `:entries` and `:meta` keys.

The `assortment` task (first discussed in https://github.com/hashobject/perun/pull/109) comes along for the ride here, because `paginate` is a subset of `assortment`.

This will probably conflict with other pull requests that I'm submitting at the same time, but they'll be trivial for me to fix as they get merged.


bhagany:
I should have mentioned - there is one backwards-incompatible change to `collection`.  The `groupby` option has been removed, because this is covered more natrually by `assortment`.  It can be added back in relatively easily if you prefer.

podviaznikov (approved):
Code looks good to me.
Only one ask would be to add use of assortment and paginate tasks into the example project https://github.com/hashobject/perun/tree/master/examples/blog

bhagany:
> Only one ask would be to add use of assortment and paginate tasks into the example project

Certainly.  I should have some time later today to do this for each of the open PR's.

podviaznikov:
ok, so 2 PRs merged and now there is a small conflict

bhagany:
Okay, I fixed this conflict.

PR #142, opened 2017-01-22
Fix #55 - Static page

### LINKED ISSUE #55: Task for rendering static pages
I'm currently doing something like this:

``` clojure
(p/collection :renderer 'org.martinklepsch.blog/error-page :page "error.html")
```

but dropping any of the passed arguments, i.e. generating a truly static html file: 

``` clojure
(defn error-page [_ _]
  (base ...))
```

While using the collection task works perfectly well in this case it might make sense to provide a general task for this usecase. Potentially task options could be used to pass arguments to the render-fn.

bhagany:
This is a simple wrapper around `render-pre-wrap` - it works exactly like `render`, except there is no `:content` key provided to the render function.  This would be easy for users to implement this themselves, but it seems to be asked for frequently enough that providing an official implementation is called for.


podviaznikov (approved):
Everything looks good. Would you mind adding this task to the example site https://github.com/hashobject/perun/tree/master/examples/blog, so that people can see how to use it.

Deraen:
This is interesting!

This doesn't fully cover my needs (or maybe it does), as I'm currently using solution where each static page has their own render function, so that they can completely control the whole page (`head` and everything). Also, I define the static pages using Clojure namespaces and I have a task that searches for namespaces under certain path etc. Anyway, this looks good, I can later see if I can port my custom solution over this.

bhagany:
> I define the static pages using Clojure namespaces and I have a task that searches for namespaces under certain path etc.

I like this a lot.  I played around with some ideas like tagging functions or namespaces with metadata for autodetection, but in the end, I went with the function symbol only because it is familiar to users of Perun.  I would be very open to ideas around expanding the capabilities of `static` like this in some way, so that it covers your use case better.  Is your custom solution available on github?

> I'm currently using solution where each static page has their own render function, so that they can completely control the whole page (head and everything).

This should be possible, but it will require having a separate `static` task for each render function, and using task order or filters to exclude the static pages from further rendering.  `static` is more like `collection` than `render` in this way, in that you have to call it once per page.

bhagany:
Hmm, I also just realized that the docstring is out of date.  I will fix this today as well.

podviaznikov:
Thanks for updating examples. Lets merge it and see how people like it.

PR #143, opened 2017-01-22
Fix #133, close #105, close #106 - decompose `markdown`

### LINKED ISSUE #105: Feature: pass pre-parse function to markdown task

### LINKED ISSUE #133: word-count task counts words in yaml header
I don't believe this would be very easy to fix in master right now, but it will be easy after a future PR.  I just wanted to leave a note about it, so I don't forget.

bhagany:
Following @MartyGentillon here: https://github.com/hashobject/perun/pull/106#issuecomment-255893969, this PR pulls `markdown` apart into two separate tasks.  The first parses the yaml header on the file, and the second only parses the markdown.  These two simpler tasks are then recomposed into a new `markdown`.

This will allow users to implement their own markdown pre/post processors with ease, and also make yaml metadata parsing easily available to future content tasks.

### DIFF
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -280,11 +280,11 @@
   "Parse markdown files
 
   This task will look for files ending with `md` or `markdown`
-  and add a `:parsed` key to their metadata containing the
-  HTML resulting from processing markdown file's content. It
-  will not parse YAML metadata at the head of the file. Also
-  writes an HTML file that contains the same content as
-  `:parsed`"
+  and writes an HTML file that contains the result from
+  processing the markdown file's content. It will _not_ parse
+  YAML metadata at the head of the file. Also adds a `:parsed`
+  key to the markdown file's metadata, allowing us not to
+  re-parse the same content later."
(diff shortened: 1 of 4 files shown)


podviaznikov (approved):
Nice, this looks great. I like how `markdown` tasks is a composition now.

PR #148, opened 2017-01-25
Adding dockerfile for example blog

podviaznikov:
Trying having dockerfile


podviaznikov:
breaks nothing, so I'll merge. If there are any comments, can apply them

PR #150, opened 2017-01-26
Update README.md

podviaznikov:
Adding @pesterhazy site to the examples section.


podviaznikov:
@bhagany ok, so now tests are also running.

PR #151, opened 2017-01-26
Add nicerthantriton.com as example

podviaznikov:
@bhagany I included your site as an example. Let me know if its ok. If ok, feel free to merge PR yourself.


bhagany:
There's not much there yet, but I don't mind :).  I'll merge when the tests are done.

As a side note, when I click on the Details link for the Runnable tests, I get redirected to their homepage.  Do you know if there's something I have to do to be able to see the details?

podviaznikov:
You can try to click signup button on Runnable and select `hashobject` org there. In that case you should be able to see logs there.

Also I work at Runnable so we have it for free for perun. If you have any issues or feedback on Runnable you can message me on slack.

bhagany:
Okay, that worked, I can see the details now.  Thanks!  +1 on free

PR #154, opened 2017-02-02
Unify `content-pre-wrap` and `render-pre-wrap`

bhagany:
`content-pre-wrap` and `render-pre-wrap` are conceptually very similar - they both read input from files (usually) and produce output files.  They are different in that `content-pre-wrap` diffs its input and only re-parses things that have changed, while `render-pre-wrap` does `clojure.tools.namespace` reloading.  By unifying the functionality of both pre-wraps, we can have render functions that don't re-render unchanged inputs, and content functions that reload when they're changed (super useful when working on Perun itself)

I added `filterer` options to the content tasks (`markdown*`, `markdown`, and `yaml-metadata`) to be consistent with the rendering tasks, and rewrote `inject-scripts` using `content-pre-wrap`.  The latter was a proof-of-concept that `content-pre-wrap` is flexible enough, because file filtering in `inject-scripts` is different from other tasks, but it still did the same kind of fileset diffing that `content-pre-wrap` does.

There is one minor backwards-incompatibility - the `:parsed`\* and `:original-path` keys have been removed. These weren't part of the public API, but their removal means that `atom-feed` and `word-count` no longer have access to the "right" content regardless of where they are called. Replicating the old behavior can be done by ensuring that `atom-feed` and `word-count` are called after `markdown` or other html-producing task, but before tasks that will modify the file content, like `render`. I think this is a more consistent and less surprising behavior, and tbh I felt like it was kind of hacky when I did it in the first place.

As always, feedback welcome.

\* `:parsed` has kind of been replaced by `:rendered`, but this key is strictly for communicating between the rendering function in a pod, and `content-pre-wrap`. It is removed from the metadata after `content-pre-wrap` is done with it.


bhagany:
I should have also mentioned that this makes it possible to have content parsing functions that operate on multiple input files and/or produce multiple output files.

bhagany:
I noted a use of `:original-path` in the perun.io repository, here: https://github.com/hashobject/perun.io/blob/master/src/io/perun/site.clj#L155.  Seems like a legit use case that this PR would break.  It would be easy enough to add it back, at least in the case of `markdown`, but I thought I'd request opinions on it before I do so.

The underlying reason I removed it (as well as `:parsed`) is that it doesn't apply to all kinds of content generally.  For instance, `collection` is backed by `content-pre-wrap`, but there is no single `:original-path`.  Similarly, `paginate` is backed by `content-pre-wrap`, and so there would be mutliple `:parsed`s.

I see two options - I could add `:original-path` and `:parsed` to the specific tasks that they make sense for, or we could require users to write their own tasks to add these kinds of keys if they want them.  Does anyone have any preferences about this, or maybe other ideas?

bhagany:
After sleeping on it, I decided to add `:original-path` back.  Still open to feedback on this.

SPEC.md
```diff
@@ -37,11 +37,6 @@ All posts have a path which is used as a key to identify the post.
 - **:content** The post content
     - Populated from file content
     - Only set in the `:entry` value of the map passed to `renderer` functions
-- **:original-path** The path for the input file from which this entry is descended
```
podviaznikov: since you added it back, can you update remove deletion from Spec?

podviaznikov (changes requested):

SPEC.md
bhagany: Certainly.  Thanks for catching that.

podviaznikov:
One comment about `:parsed`. Before it was possible to see in `print-meta` what was the :parsed markdown. Now it wouldn't be possible.
I think we could overcome this in the future by having an option to print content of the file itself (in `print-meta`), right?

bhagany:
Yes, that's true.  I like the idea of an option in `print-meta` to see the content - I'll add it to this PR.

podviaznikov (approved):

podviaznikov:
Thank you!

PR #155, opened 2017-02-02
Add 200ok.ch as a Perun example page

munen:
If you're interested in getting one more page on the list that looks a little different under the hood than the existing examples. Of course it's totally OK to reject the PR, because you already have a couple good examples of Perun pages. I won't be offended(;


podviaznikov:
Absolutely, real world examples are perfect. Thank you!

PR #161, opened 2017-02-13
Provide default for `print-meta`'s `content-exts`

bhagany:
Calling `print-meta` without passing `content-exts` previously
caused a NullPointerException.


bhagany:
This is a simple bugfix, and tests pass, so I'm going to merge this one.

PR #163, opened 2017-02-14
Atom upgrades

bhagany:
My primary motivation was to get paginated Atom feeds, so that more than 10 items can be retrieved by clients, but also feed pages don't get too large. Along the way, I also addressed some of the FIXME's in atom.clj, upgraded the error reporting, and made good use of `content-pre-wrap`.

Enabling pagination for `atom-feed` entailed upgrading `paginate` as well - it now adds more useful metadata to the generated pages.  I've updated the spec document accordingly.

There is one backwards incompatible change for `paginate`.  Where it previously took a `prefix` option, it now takes a `filename-fn`.  This can be made to do the same thing `prefix` did, but also allows you to come up with page naming schemes that don't fall within the narrow confines of page-1, page-2, etc.

### DIFF
--- a/src/io/perun/meta.clj
+++ b/src/io/perun/meta.clj
@@ -10,7 +10,8 @@
 (def +global-meta-key+ :io.perun.global)
 
 (def +global-meta-defaults+
-  {:doc-root "public"})
+  {:doc-root "public"
+   :io.perun/version perun/+version+})
 
 (defn get-global-meta
   "Return global metadata that is related to the whole project
--- a/src/io/perun/meta.clj
+++ b/src/io/perun/meta.clj
@@ -29,11 +29,7 @@
   (let [file (or file (io/file path))
         filename (.getName file)
         slug (slug filename)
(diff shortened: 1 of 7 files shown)


bhagany:
sigh.  I thought this would merge cleanly.  I'll fix when I'm able.

src/io/perun/meta.clj
```diff
@@ -10,7 +10,8 @@
 (def +global-meta-key+ :io.perun.global)
 
 (def +global-meta-defaults+
-  {:doc-root "public"})
+  {:doc-root "public"
+   :io.perun/version perun/+version+})
```
podviaznikov, line 14: this is nice
👍

podviaznikov (approved):

bhagany:
While merging from master, I changed my mind a bit about `paginate`.  Rather than having a `filename-fn`, I think it's more consistent with other tasks to have a `slug-fn`, plus `out-ext` for determining filenames.

PR #167, opened 2017-02-18
Fix #165 - Allow multiple `update!` calls in the same render pod

### LINKED ISSUE #165: New static task is not hot-reloading for me
I may just be doing something wrong, my page built with `static` does not hot reload while other changes with `collection` or `render` reload fine.

You can reproduce in this repo: https://github.com/sooheon/sooheon.com/blob/master/src/site/layout.clj#L68

bhagany:
All tasks that have a `renderer` option use the same pod, for
efficiency. Because of this, when multiple `renderer` tasks were
used in the same task chain, only the first one would pick up on
namespace refreshes, which meant that things that ought to be
re-rendered would not be.

This change causes each pod to track whether it has any refreshed
namespaces on this task loop, and to return the correct status, even
if the refreshing had actually be done by a task further up in the
chain. Then, after the chain runs, the atom that tracks each pod's
status is reset.

Due to this change the names `content-pre-wrap`, `render-pre-wrap`,
and `assortment-pre-wrap` are no longer accurate, because the render
pod's atom is reset after subsequent tasks run. They have been
changed to `content-task`, `render-task`, and `assortment-task`.

### DIFF
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -220,7 +220,7 @@
 (def ^:private content-deps
   '[[org.clojure/tools.namespace "0.3.0-alpha3"]])
 
-(defn content-pre-wrap
+(defn content-task
   "Wrapper for input parsing tasks. Calls the return from `render-form-fn` on
   new or changed inputs returned from `paths-fn`, adds `tracer` to
   `:io.perun/trace` and writes files for subsequent tasks to process, if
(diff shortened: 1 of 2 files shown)


bhagany:
No API changes here, and it fixes an annoying bug, so I'm merging.

PR #169, opened 2017-02-18
Lucene word count

bhagany:
Previously, word counting just split on whitespace and counted the parts, but this doesn't really count words for the common use-cases in Perun.  For instance, if you run `word-count` on this markdown:

```markdown
# Hello there

This --- be ___markdown___.
```

it counts 8 words, but it's really 5.  It's worse if you run `word-count` on HTML, because it would count things that are inside tags like `<a href="/" class="some-class">`.

This PR fixes all that by using Lucene's standard tokenizer, which should work for most languages.  It won't work that great for Chinese, Japanese, or Thai, but splitting on whitespace won't work for those languages either, and I believe support can be added to this implementation later if there's a demand.

### DIFF
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -445,7 +445,7 @@
         (perun/report-debug "ttr" "generated time-to-read" (map :ttr updated-metas))
         (pm/set-meta fileset updated-metas)))))
 
-(def ^:private word-count-deps
+(def ^:private ^:deps word-count-deps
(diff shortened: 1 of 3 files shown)


podviaznikov:
This looks definitely much better than before. Good idea with using Lucene for this task

bhagany:
Thanks :)  now if I can just merge things correctly

PR #170, opened 2017-02-19
Fix #126 - replace Pegdown with flexmark-java

### LINKED ISSUE #126: Pegdown has been deprecated; pick a new markdown parser
I just noticed that Pegdown has been EOL'd as of five days ago (see README: https://github.com/sirthias/pegdown), so I think a replacement is probably worth it for Perun.  I've been toying with the idea of adding Pandoc support, and there are a few other mentions of this in other issues, but I'm not sure if Pandoc can support all of the same features that Pegdown does.  Pegdown's deprecation notice recommends switching to [flexmark-java](https://github.com/vsch/flexmark-java), which is apparently a drop-in replacement.

I can integrate one or both of these, but I'd like to get some opinions on what you'd like to see, whether Pandoc, flexmark-java, some other option, or possibly multiple integrations.  Let me know what you think.

bhagany:
flexmark-java is now at feature parity with Pegdown, so there are no more roadblocks to replacing it.

No API changes here, though there is at least one minor parsing difference between the two, and maybe more.  Should probably test on a few projects before merging (I've tested three that seem fine).


podviaznikov (approved):

podviaznikov:
Good job!

PR #172, opened 2017-02-19
make `content-task` return the fileset

bhagany:
Thanks to @martinklepsch for spotting my silly mistake here


bhagany:
hrm, Runnable keeps failing, but never for the same reason.  I think the tests are taking too long to complete, and also, the last one ran out of memory.

I had an idea about making this faster the other day, I guess I'll give it a try.

Issue #48, opened 2015-07-13
Allow markdown task to use an external markdown parser


greywolve:
I've added this in a fork over [here](https://github.com/greywolve/perun/blob/master/src/io/perun.clj#L36). I wanted this so I could use [Pandoc](http://www.pandoc.org) instead of Endophile for markdown parsing. Would you be interested in supporting this ? 

``` clj
;; argument is an edn vector of strings
(markdown :cmd ["pandoc"])
```

Thanks for the library by the way, really cool idea.

Deraen:
This is definitively an interesting idea. Pandoc can also do code highlighting (#32).

I wonder if an separate task and maybe library would be good idea for this?

Separate task might be good choice as it would be useful to define option for Pandoc and we probably don't want to make markdown task too complicated.

greywolve:
I was going to make an isolated task, but then I realized it would be basically reusing all the same code that already exists, barring a few small changes. Though you are probably right, it makes more sense as a separate task. I'd be happy to make one (though there will be a fair bit of duplication).

Deraen:
Once it's separated as a task we could probably move the shared logic to functions.

Deraen:
Might also be useful to create Clojure wrapper for Pandoc, like: https://github.com/pvorb/scala-pandoc

Mostly it would just translate given options to cli arguments, and take care of calling the command properly.

greywolve:
Great idea. :) I'll get onto this.

Deraen:
And note about using `conch`, you shouldn't add transitive dependencies to the perun and instead you should add the dependencies to just the pods.

But as it happens, boot is already using `conch`: https://github.com/boot-clj/boot/blob/master/boot/pod/src/boot/from/me/raynes/conch.clj so you can use `boot.from.me.raynes.conch` ns in your code without adding any dependencies.

... Though if Pandoc wrapper is created, that's what would be using Conch and it would be used through a pod.

greywolve:
Ahh, I knew boot was using `conch` but wasn't aware they actually had it copied to a namespace inside boot, good to know!

podviaznikov:
@greywolve it's a great idea to add pandoc support. I think it would be useful.
I usually check metalsmith for the inspirations (http://www.metalsmith.io/), they also have `pandoc` as separate task with task specific options: https://github.com/arve0/metalsmith-pandoc.
So I also agree with @Deraen suggestion to make separate task.

Deraen:
[...]

PR #173, opened 2017-02-21
Add pandoc task

bhagany:
Seeking feedback on this - Pandoc has been discussed before (notably in #48), with a few different approaches presented.  The current state of this PR is "the simplest thing possible" - the `pandoc` task takes a vector of command line arguments to pass on to the `pandoc` binary.  I'm not sure this is the best way, and I'd like to hear what the community thinks.

### DIFF
--- a/Dockerfile_test
+++ b/Dockerfile_test
@@ -6,7 +6,7 @@ ENV DEBIAN_FRONTEND noninteractive
 # Oracle Java 8
 
 RUN apt-get update \
-    && apt-get install -y curl wget openssl ca-certificates \
+    && apt-get install -y curl wget openssl ca-certificates pandoc\
     && cd /tmp \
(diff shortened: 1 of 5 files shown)


bhagany:
Also the tests say they pass, but it's not really true because the `pandoc` binary isn't installed in the container.  I'm unsure how to do that...

podviaznikov:
Yeah, I was surprised that tests are green even if they failed.
To install it in the container https://github.com/hashobject/perun/blob/master/Dockerfile_test should be modified to install Pandoc.

podviaznikov:
trying to add pandoc myself to the dockerfile

podviaznikov:
ok, so tests properly passed this time

podviaznikov (approved):

podviaznikov:
This makes sense to me. Shows that content tasks can be different.

Issue #181, opened 2017-03-15
Add HTML Manifest support


podviaznikov:
We can create task that will generate HML manifest. Like this one (https://podviaznikov.com/manifest.json).
This will allow Perun site behave more like an app.
This will work great together with https://github.com/hashobject/perun/issues/180

Issue #180, opened 2017-03-15
Add support for ServiceWorkers (a.k.a. "offline mode")


podviaznikov:
There is new HTML feature called `ServiceWorkers`. This allows to cache all needed resources and make site work offline without Internet connection.

This seems like a good feature for blogs, which can be opened once and read later.

I implemented such caching service worker for my personal site: https://podviaznikov.com/sw.js.
It would be good to have something similar as perun task.
GatsbyJS has such support built in. We can do something similar for sure: https://github.com/gatsbyjs/gatsby/issues/670

PR #182, opened 2017-03-26
Implement #181 - web app manifest support

bhagany:
This feature will allow sites built with Perun to be installed as an app on systems that support it. When paired with #180, this will be a slick offline experience.

This was a bit more work to implement properly than I thought at first, primarily because I wanted to support automatically handling image files for the icons that are referenced in manifest.json.  This required a refactor for `images-resize`, which is backwards-compatible.  This in turn caused another small backwards-compatible addition to `content-task` to allow a boot tmp directory to be passed.  This was necessary to accommodate tasks that want to write directly to tmp directories from within their pod, rather than passing data back to the main pod, as is the case for the image-handling tasks.

Other changes here are small improvements like pruning unused requires, and consistently handling the `out-dir` parameter to tasks.  These should be self-explanatory, but if you have questions let me know.

I've also added tests for `manifest`.

### DIFF
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -1285,3 +1285,61 @@
         :passthru-fn content-passthru
         :task-name "inject-scripts"
         :tracer :io.perun/inject-scripts}))))
+
+(def ^:private ^:deps manifest-deps
+  '[[org.clojure/tools.namespace "0.3.0-alpha3"]
+    [cheshire "5.7.0"]])
+
+(def +manifest-defaults+
+  {:out-dir "public"
+   :icon-path "icon.png"
+   :resolutions #{192 512}
+   :theme-color "#ffffff"
+   :display "standalone"
+   :scope "/"})
+
+(deftask manifest*
+  [o out-dir     OUTDIR  str "the output directory"
+   t site-title  TITLE   str "name for the installable web application"
+   c theme-color COLOR   str "background color theme for icon (default \"#ffffff\")"
+   d display     DISPLAY str "display mode for browser (default \"standalone\")"
+   s scope       SCOPE   str "the scope to which the manifest applies (default \"/\")"]
+  (let [{:keys [site-title] :as opts} (merge +manifest-defaults+ *opts*)
+        pod (create-pod manifest-deps)]
+    (letfn [(manifest-path [fileset]
+              (let [icon-metas (filter-meta-by-ext fileset {:filterer :manifest-icon})
+                    path (perun/create-filepath out-dir "manifest.json")
+                    global-meta (pm/get-global-meta fileset)
+                    args (merge opts
+                                {:icons icon-metas
+                                 :input-paths (into #{} (map :path icon-metas))
+                                 :site-title (or site-title (:site-title global-meta))})]
+                {path args}))]
+      (content-task
+       {:render-form-fn (fn [data] `(io.perun.manifest/manifest ~data))
+        :paths-fn manifest-path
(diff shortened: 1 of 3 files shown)


bhagany:
I should also mention my sources for how to work with a manifest.  The spec is here: https://w3c.github.io/manifest/, and this was useful for seeing how it's actually supported in browsers: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json.  I paid the most attention to the output from http://realfavicongenerator.net/, which does a nice job of exposing the most relevant parts of a manifest.  It seems to focus primarily on Android, preferring a different approach for other operating systems, hence the docstring for `manifest`.  I plan on exploring this further, but this PR is a good base to build from.

src/io/perun.clj
```diff
@@ -1167,3 +1210,61 @@
...
+                {path args}))]
+      (content-task
+       {:render-form-fn (fn [data] `(io.perun.manifest/manifest ~data))
+        :paths-fn manifest-path
+        :task-name "manifest"
+        :tracer :io.perun/manifest
+        :pod pod}))))
+
+(deftask manifest
+  "Creates a manifest.json for Android (currently)"
```
podviaznikov: why it's only for android? Wouldn't it work also on the desktop? What makes it Android specific?

src/io/perun/contrib/images_resize.clj
```diff
@@ -9,29 +9,29 @@
...
-
-(defn resize-to [tgt-path file options resolution]
-  (let [io-file (-> file :full-path io/file)
-        buffered-image (iu/buffered-image io-file)
-        resized-buffered-image (resize/resize-to-width buffered-image resolution)
-        new-dimensions (iu/dimensions resized-buffered-image)
-        new-meta (write-file options tgt-path file resized-buffered-image resolution)
-        dimensions {:width (first new-dimensions) :height (second new-dimensions)}]
-    (merge file new-meta dimensions (select-keys options [:out-dir]))))
+(def img-cache (atom {}))
```
podviaznikov: oh, that is nice

podviaznikov (changes requested):

src/io/perun.clj
bhagany: I definitely don't understand everything about this, so I may be completely wrong. But for what it's worth - I took my cue from the Real Favicon Generator (http://realfavicongenerator.net/), which only references icons specifically for Android in manifest.json.  I also can't find any examples of how manifest.json is supported on the desktop, and according to https://developer.mozilla.org/en-US/docs/Web/Manifest, no desktop browsers support it.

src/io/perun/contrib/images_resize.clj
bhagany: Glad you like it :)

bhagany:
I've split this PR into a few more parts (#203 and #205) so that they can be considered individually.  I will also be updating this PR soon to improve functionality, and address the comments.

bhagany:
@podviaznikov I've made a few improvements, and I believe I've addressed your comment on the docstring (you were right).  Let me know if you'd like to see other changes.

arichiardi:
This is a great feature!

podviaznikov (approved):
Good job!
If you will have time, please update one of our example projects to use it. If you don't have time, merge away and create a ticket in the repo (to update example project with this task). Maybe someone else would be able to contribute it

allentiak:
Hi, @bhagany 

Thanks again for your hard work!
Are you still working on this?

PR #183, opened 2017-04-09
feat: asciidoctor plugin

nicorikken:
A basic tasks for processing asciidoc files using AsciidoctorJ. This implementation is bare-bones, it does not cover Asciidoctor extensions and does not relate asciidoc attributes with Perun meta.
The page title is not rendered by the Asciidoctor task, to mimic the behavior of the markdown extension.

### DIFF
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -458,6 +458,55 @@
                    :meta meta
                    :cmd-opts cmd-opts))))
 
+(def ^:private ^:deps asciidoctor-deps
+  '[[org.clojure/tools.namespace "0.3.0-alpha3"]
+    [org.asciidoctor/asciidoctorj "1.5.4"]])
+
+(def ^:private +asciidoctor-defaults+
+  {:out-dir    "public"
+   :out-ext    ".html"
+   :filterer   identity
+   :extensions [".ad" ".asc" ".adoc" ".asciidoc"]
+   :meta       {:original     true
+                :include-rss  true
+                :include-atom true}})
+
+(deftask asciidoctor*
+  "Parse asciidoc files using Asciidoctor
+
+   Asciidoctor has basic support for markdown, and can therefore also be used
+   for parsing `.md` files."
+  [d out-dir    OUTDIR     str   "the output directory"
+   _ filterer   FILTER     code  "predicate to use for selecting entries (default: `identity`)"
+   e extensions EXTENSIONS [str] "extensions of files to process"
+   m meta       META       edn   "metadata to set on each entry"]
+  (let [pod     (create-pod asciidoctor-deps)
+        options (merge +asciidoctor-defaults+ *opts*)]
+    (content-task
+     {:render-form-fn (fn [data] `(io.perun.asciidoctor/process-asciidoctor ~data))
+      :paths-fn       #(content-paths % options)
+      :passthru-fn    content-passthru
+      :task-name      "asciidoctor"
+      :tracer         :io.perun/asciidoctor
+      :rm-originals   true
+      :pod            pod})))
+
+(deftask asciidoctor
+  "Parse asciidoc files with yaml front matter using Asciidoctor
+
+   Asciidoctor has basic support for markdown, and can therefore also be used
+   for parsing `.md` files."
+  [d out-dir    OUTDIR     str   "the output directory"
+   _ filterer   FILTER     code  "predicate to use for selecting entries (default: `identity`)"
+   e extensions EXTENSIONS [str] "extensions of files to process"
+   m meta       META       edn   "metadata to set on each entry"]
+  (let [{:keys [out-dir filterer extensions meta]} (merge +asciidoctor-defaults+ *opts*)]
+    (comp (yaml-metadata :filterer filterer :extensions extensions)
+          (asciidoctor* :out-dir    out-dir
+                        :filterer   filterer
+                        :extensions extensions
+                        :meta       meta))))
+
 (deftask global-metadata
   "Read global metadata from `perun.base.edn` or configured file.
 
--- a/src/io/perun/asciidoctor.clj
+++ b/src/io/perun/asciidoctor.clj
@@ -1,0 +1,13 @@
+(ns io.perun.asciidoctor
+  (:require [io.perun.core :as perun]
+            [clojure.java.io :as io])
+  (:import [org.asciidoctor Asciidoctor Asciidoctor$Factory]))
+
+(defn asciidoctor-to-html [file-content]
+  (.convert (Asciidoctor$Factory/create "") file-content {}))
+
+(defn process-asciidoctor [{:keys [entry]}]
+  (perun/report-debug "asciidoctor" "processing asciidoctor" (:filename entry))
+  (let [file-content (-> entry :full-path io/file slurp)
+        html         (asciidoctor-to-html file-content)]
+    (assoc entry :rendered html)))
--- a/src/io/perun/asciidoctor.clj
+++ b/src/io/perun/asciidoctor.clj
@@ -3,8 +3,11 @@
             [clojure.java.io :as io])
   (:import [org.asciidoctor Asciidoctor Asciidoctor$Factory]))
 
+(def container
+  (Asciidoctor$Factory/create ""))
+
 (defn asciidoctor-to-html [file-content]
-  (.convert (Asciidoctor$Factory/create "") file-content {}))
+  (.convert container file-content {}))
 
 (defn process-asciidoctor [{:keys [entry]}]
   (perun/report-debug "asciidoctor" "processing asciidoctor" (:filename entry))
--- a/test/io/perun_test.clj
+++ b/test/io/perun_test.clj
@@ -188,11 +188,22 @@
 
 This --- be ___markdown___.")
 
+(def adoc-content
+  "= Hello there
+
+This --- be _asciidoc_.")
+
 (def input-strings (map #(str "---\n" % "\n---\n" md-content) yamls))
+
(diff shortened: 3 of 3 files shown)


nicorikken:
For the backstory of this PR, please read https://github.com/hashobject/perun/pull/100

nicorikken:
Any feedback?

bhagany:
Hi Nico!  I'm so glad you've found time to work on this.  I apologize - you've caught me in the middle of a job transition (today is my second day) and I haven't had time to do a proper review.  I should be able to do so this weekend, though.  Sorry for the delay!

nicorikken:
@bhagany Thanks for letting me know. I'm excited to have something mergeable, so I'm eager to get it reviewed, that's all. But certainly it can wait some more, no problem.

podviaznikov (commented):
code looks good to me. Waiting for @bhagany review before merging

src/io/perun.clj
```diff
@@ -458,6 +458,55 @@
...
+  "Parse asciidoc files with yaml front matter using Asciidoctor
+
+   Asciidoctor has basic support for markdown, and can therefore also be used
+   for parsing `.md` files."
+  [d out-dir    OUTDIR     str   "the output directory"
+   _ filterer   FILTER     code  "predicate to use for selecting entries (default: `identity`)"
+   e extensions EXTENSIONS [str] "extensions of files to process"
+   m meta       META       edn   "metadata to set on each entry"]
+  (let [{:keys [out-dir filterer extensions meta]} (merge +asciidoctor-defaults+ *opts*)]
+    (comp (yaml-metadata :filterer filterer :extensions extensions)
```
bhagany, line 504: Since Asciidoc can use YAML front-matter itself, you might consider passing `:keep-yaml true` to `yaml-metadata`.  That way, the YAML will still be present in the file when the `asciidoctor*` task sees it.

src/io/perun/asciidoctor.clj
```diff
@@ -0,0 +1,13 @@
+(ns io.perun.asciidoctor
+  (:require [io.perun.core :as perun]
+            [clojure.java.io :as io])
+  (:import [org.asciidoctor Asciidoctor Asciidoctor$Factory]))
+
+(defn asciidoctor-to-html [file-content]
+  (.convert (Asciidoctor$Factory/create "") file-content {}))
```
bhagany: This might not make much difference, but rather than creating the Asciidoctor interface on each call to `asciidoctor-to-html`, would it be more efficient to say, `(def asciidcotor (Asciidoctor$Factory/create ""))` as a top-level form, and then reuse the same object?

test/io/perun_test.clj
```diff
@@ -400,6 +425,26 @@ This --- be ___markdown___.")
                          :msg "`pandoc` should parse HTML to markdown"))
 
         (add-txt-file :path "test.md" :content (nth input-strings 0))
+        (p/asciidoctor :out-dir nil
```
bhagany, line 428: I like that you're testing different parsing modes

bhagany:
I'm very pleased with how clean this implementation is.  I think it's a good base on which more Asciidoctor features can be added over time.  Thanks!

src/io/perun.clj
nicorikken: I agree, but the issue arises that the default meta keys used in Perun and !{Asciidoctor](http://asciidoctor.org/docs/user-manual/#attribute-catalog) are different. I was thinking of introducing a translation layer in a next version which contains the necessary translation layer, as I stated in the currently last comment on https://github.com/hashobject/perun/pull/100 For now I could indeed also just pass it through. What do you think, considering my comments? Both are fine to me.

src/io/perun/asciidoctor.clj
nicorikken: Yes, but this fails the moment that environment-specific parameters are introduced like libraries. The issue of creating new containers came up on https://github.com/hashobject/perun/pull/100 and does hurt performance for large number of conversions.
I guess not creating new containers does limit the conversion process to a single core?
Probably the converter-specific settings have to be raised to the task level, so that a different container setting requires a different task to be created. Then inside the task the same converter (factory) can be used.
I will refactor this.

test/io/perun_test.clj
nicorikken: Then again I couldn't get the tests to fail based on content, so perhaps I'm using the framework incorrectly.

nicorikken:
@bhagany Let me know what you think about the options for dealing with YAML, and if I'm missing out om some testing features in the testing framework. I'll refactor the initiation of new JRuby containers.

src/io/perun/asciidoctor.clj
```diff
@@ -0,0 +1,13 @@
+(ns io.perun.asciidoctor
```
MartyGentillon, line 1: See above.

src/io/perun/asciidoctor.clj
```diff
@@ -0,0 +1,13 @@
...
+  (:require [io.perun.core :as perun]
+            [clojure.java.io :as io])
+  (:import [org.asciidoctor Asciidoctor Asciidoctor$Factory]))
+
+(defn asciidoctor-to-html [file-content]
+  (.convert (Asciidoctor$Factory/create "") file-content {}))
+
+(defn process-asciidoctor [{:keys [entry]}]
+  (perun/report-debug "asciidoctor" "processing asciidoctor" (:filename entry))
```
MartyGentillon: See above.

src/io/perun.clj
```diff
@@ -458,6 +458,55 @@
...
+    (content-task
+     {:render-form-fn (fn [data] `(io.perun.asciidoctor/process-asciidoctor ~data))
+      :paths-fn       #(content-paths % options)
+      :passthru-fn    content-passthru
+      :task-name      "asciidoctor"
+      :tracer         :io.perun/asciidoctor
+      :rm-originals   true
+      :pod            pod})))
+
+(deftask asciidoctor
```
MartyGentillon, line 494: By naming the task after the library you are using, you bind the implementation to the interface. A better task name would be ascidoc.

MartyGentillon (commented):
Nice work.  A few thoughts.

nicorikken:
@MartyGentillon I see your point of view, but then again, there is also an outdated [Python reference implementation of Asciidoc](http://www.methods.co.nz/asciidoc/) which is quite outdated  Small parts of syntax in Asciidoctor are still improving, to end up at the end-goal of an 'unidoc' syntax. Also Asciidoctor has some Markdown support, and has the ability to load some plugins, like the terrific [asciidoctor-diagram](https://github.com/asciidoctor/asciidoctor-diagram). So while I see where you are coming from, I'd rather leave it at the tool, for clarity.

MartyGentillon:
That seems like a fairly good argument to me.  I wonder if it makes sense
to add aliases for those who don't care what Asciidoc interpreter they
use.  It would give the best of both worlds, people woo don't care won't be
broken by a parser change, while those who do can use the advanced features.

nicorikken:
- The change to share the Asciidoctor JRuby container is committed.
- I tried to handle the YAML in Asciidoc, but could not manage to. So the only option would be to skip it altogether. Also I think keeping the same pattern as the Markdown task benefits the project structure. I consider exposing YAML metadata to the Asciidoctor tasks a desirable next step.
- I asked about the tests, which do not seem to test on content. Perhaps this can be improved at a later stage.

To me the code is ready to merge in it's current state. Metadata would be the focus for the next iteration.

bhagany:
Seems to already be decided, but because this task can parse different types of content, it seems to be closer to `pandoc` (which is named after the tool), than it is to `markdown`.  So, I agree that `asciidoctor` is a more appropriate name for the task.

I do want to take a look at the tests, and I should have time tonight or tomorrow, but otherwise I agree this is ready to merge.  I'm waiting for the legal team at my new job to sign off on my work on Perun, so I don't want to merge before that happens.  If one of the other committers wants to after we figure out what's up with the tests though, I approve.

bhagany:
Using aliases based on the type of content to be parsed is an interesting idea, and I think it deserves more thought.  There would be quite a few aliases for `pandoc`, but we'd probably be able to at least do the most popular ones.

nicorikken:
@bhagany Thanks for the update. If somebody can verify I implemented the tests correctly, that would be great.

bhagany:
@nicorikken I think the tests are implemented correctly.  I was able to get them to fail by changing the `parsed-asciidoc-md` and `parsed-asciidoc-adoc` vars to something else.  I might be missing something though, could you explain in more detail how you got them to behave in a way you didn't expect?

nicorikken:
I was expecting a summary of the test results, as I know from my Leiningen usage. But indeed I get error output if the testresults are incorrect:
```
[asciidoctor] - rendered new or changed file test.html

FAIL in clojure.lang.PersistentList$EmptyList@1 (perun_test.clj:59)
io.perun-test/with-arguments-test
`asciidoctor` should populate HTML with parsed markdown content
expected: (f (.contains (slurp (boot/tmp-file (boot/tmp-get fileset path))) content))
  actual: false
```
So in hindsight it was just my expectation, not the actual code.

bhagany:
Okay, then I think this is ready to merge.  Still haven't heard back from my legal department, but if another committer wants to merge before that happens, I'm good with that.

MartyGentillon (approved):

nicorikken:
Considering that all participants on this PR approved, can this be merged?

bhagany:
I got permission from the legal department :)

nicorikken:
Nice!
Finally getting Asciidoctor is a relief. Thanks!

PR #191, opened 2017-08-20
feat(asciidoctor): handle metadata

nicorikken:
Adds metadata handling to the asciidoctor task. The asciidoctor task will now:
1. read asciidoctor attributes and add them as metadata
2. expose metadata to the asciidoctor process as attributes

Both processes are executed independently from each other.

### DIFF
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -459,7 +459,8 @@
                    :cmd-opts cmd-opts))))
 
 (def ^:private ^:deps asciidoctor-deps
-  '[[org.clojure/tools.namespace "0.3.0-alpha3"]
+  '[[clj-time "0.14.0"]
+    [org.clojure/tools.namespace "0.3.0-alpha3"]
     [org.asciidoctor/asciidoctorj "1.5.4"]])
 
 (def ^:private +asciidoctor-defaults+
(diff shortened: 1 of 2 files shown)


nicorikken:
As promised (https://github.com/hashobject/perun/pull/100), a new Asciidoctor pull request, for handling metadata.

Some notes for review:
- I added s protective function to prevent the asciidoc attributes from messing with the Perun framework (`protect-meta`). Am I missing some important keywords here?
- I wasn't sure what metadata features I was to enable to the task options. Now you can opt for the `asciidoctor*` task instead of the `asciidoctor` task in order to leave out the yaml frontmatter parsing. But regardless the document attributes are exposed to Perun, and the Perun metadata is exposed to the asciidoctor conversion. There is a multitude of options here, so please shine your light on what makes sense.
- This does not include any tests. I did do some tests on my local system though. Ideally I would make some small unit-tests in a separate file, but I wasn't able to add a separate file to the testing process. Perhaps I have to looker further into the workings of Boot.

nicorikken:
As mentioned I'm also working on getting asciidoctor-diagram to function. I'm aiming for the inline svg solution (https://github.com/asciidoctor/asciidoctor/issues/1301) as a start, not having to deal with the Boot file handing. Somehow the JRuby container cannot access the binaries on my system for the actual image generation, so that'll take some more time. Similarly perhaps syntax highlighting will become a feature.

podviaznikov (approved):

allentiak (changes requested):
Hi, @nicorikken 

Thanks for your hard work! Maybe this PR has never been merged because it lacks tests?
You may find useful links in #228.

Issue #192, opened 2017-08-20
Yaml parsing fails on asciidoc code block


nicorikken:
The `parse-yaml` task fails when no YAML frontmatter is provided, but a valid asciidoctor codeblock is encountered further down the file:

Minimal source example:
```
= Bugreport

[source, json]
----
{"a", "b"}
----
```

Stacktrace:
```
org.yaml.snakeyaml.parser.ParserException: null; expected '<document start>', but found BlockSequenceStart;  in 'string', line 2, column 1:
                                               -
                                               ^
        problem: "expected '<document start>', but found BlockSequenceStart"
    problemMark: #object[org.yaml.snakeyaml.error.Mark 0x7d787635 " in 'string', line 2, column 1:\n    -\n    ^"]
org.yaml.snakeyaml.parser.ParserImpl$ParseDocumentStart.produce       ParserImpl.java:  225
                 org.yaml.snakeyaml.parser.ParserImpl.peekEvent       ParserImpl.java:  158
                org.yaml.snakeyaml.parser.ParserImpl.checkEvent       ParserImpl.java:  143
             org.yaml.snakeyaml.composer.Composer.getSingleNode         Composer.java:  108
   org.yaml.snakeyaml.constructor.BaseConstructor.getSingleData  BaseConstructor.java:  120
                         org.yaml.snakeyaml.Yaml.loadFromReader             Yaml.java:  481
                                   org.yaml.snakeyaml.Yaml.load             Yaml.java:  400
                                                            ...                            
                                     clj-yaml.core/parse-string              core.clj:  120
                                                            ...                            
                                       io.perun.yaml/parse-yaml              yaml.clj:   48
                                                            ...                            
                                             clojure.core/apply              core.clj:  630
                                          boot.pod/eval-fn-call               pod.clj:  328
                                              boot.pod/call-in*               pod.clj:  379
                                                            ...                            
                                              boot.pod/call-in*               pod.clj:  382
                                 io.perun/render-in-pod/iter/fn             perun.clj:  175
                                                            ...                            
                                               clojure.core/seq              core.clj:  137
                                            clojure.core/map/fn              core.clj: 2616
                                                            ...                            
                                              clojure.core/next              core.clj:   64
                                 io.perun/content-task/fn/fn/fn             perun.clj:  257
                                    io.perun/content-task/fn/fn             perun.clj:  257
                                      io.perun/eval725/fn/fn/fn             perun.clj:  519
[...]

PR #193, opened 2017-08-28
fix: asciidoc yaml parsing

nicorikken:
Fixes the issue as described in https://github.com/hashobject/perun/issues/192
by checking the triple dash to be prefixed by either the start of the document,
or the start of a new line. This does still allow for some content before the
actual yaml.

### DIFF
--- a/src/io/perun/yaml.clj
+++ b/src/io/perun/yaml.clj
@@ -7,7 +7,7 @@
   (:import [flatland.ordered.map OrderedMap]
            [flatland.ordered.set OrderedSet]))
 
-(def ^:dynamic *yaml-head* #"---\r?\n")
+(def ^:dynamic *yaml-head* #"(?<=^|\n)---\r?\n")
 
 (defn substr-between
   "Find string that is nested in between two strings. Return first match.


nicorikken:
I just changed the PR to use lookbehind rather than a match, as the match could not cope with a the two `yaml-head`'s coming right after each other. Both `yaml-head`'s were wanting to consume the single newline inbetween. The backtracking solves this issue.

nicorikken:
Again I just changed the PR, as I made a typo in the lookbehind syntax, causing false matches.

podviaznikov (approved):

podviaznikov:
this will be tricky one. We need to test this somehow. High risk it might break people's sites.

podviaznikov:
maybe we can include couple of tests?
And also add this test case https://github.com/hashobject/perun/issues/192 to the tests?

allentiak:
Hi, @nicorikken 

Still working on this? Any news regarding @podviaznikov's suggestion?

allentiak (changes requested):
(Just to make @podviaznikov's feedback explicit.)

PR #194, opened 2017-09-07
feat: asciidoctor-diagram support

nicorikken:
Enables `asciidoctor-diagram` support for generating images as defined inline in the source file. The resulting images are captured afterwards in a new task named `collect-images`, to stick to the basic content-task structure for the base asciidoctor task.

Previous code was refactored to have stronger defaults for options and metadata handling.

Based on the previous PR's https://github.com/hashobject/perun/pull/191 and https://github.com/hashobject/perun/pull/193

I'm very interested to hear your comments on this work:
- Is the `collect-images` task the right way to collect images after the `content-task` function?
- The `asciidoctor*` now has a global switch on `diagram` to determine whether or not `asciidoctor-diagram` is enabled. Is this correct?
- Is there a way I can move both the images and the post with the `permalink` task so the relative URL remains the same? Another way would be to fix this upfront, by giving each post it's own directory or by changing the relative URL named `imagesdir` in asciidoctor terminology.

At the moment of writing you can observe the potential output at http://site.nicorikken.eu/posts/2017-08-14-testpost.html which is based on the source https://github.com/nicorikken/nicorikken.github.io/tree/perun which uses the code of this PR.

### DIFF
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -459,7 +459,8 @@
                    :cmd-opts cmd-opts))))
 
 (def ^:private ^:deps asciidoctor-deps
-  '[[org.clojure/tools.namespace "0.3.0-alpha3"]
+  '[[clj-time "0.14.0"]
+    [org.clojure/tools.namespace "0.3.0-alpha3"]
     [org.asciidoctor/asciidoctorj "1.5.4"]])
 
 (def ^:private +asciidoctor-defaults+
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -9,7 +9,8 @@
             [clojure.string :as string]
             [clojure.edn :as edn]
             [io.perun.core :as perun]
-            [io.perun.meta :as pm]))
+            [io.perun.meta :as pm]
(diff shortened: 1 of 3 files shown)


nicorikken:
I hope somebody will take a look at this PR, as it feels like a significant milestone to me.

bhagany:
Hi Nico! Sorry for the slow responses lately. I'm trying to ramp back up on Perun after an unplanned break due to starting a new job.  I'd like to get a handle on your asciidoctor PR's so that we can get them merged, but I'm a bit confused on where I should start to get an overall idea of what's going on. Can you point me in a direction?

podviaznikov (commented):
Code looks good to me. Would you mind adding example site into the examples directory so that 1) people can see how to use it, 2) we can test it

allentiak:
Hi, @nicorikken!
Still working on this? What do you think about @podviaznikov's suggestion?

allentiak (changes requested):
(Just to make @podviaznikov's feedback explicit.)

PR #201, opened 2017-12-11
Protect in case of missing <pre> in highlight

arichiardi:
The patch adds a some-> so that we avoid NullPointerExceptions if <pre> tags
are missing in the posts.


bhagany:
Thanks!

Issue #199, opened 2017-11-25
Permalink wrong with out-dir?


arichiardi:
Hello folks!

I am rendering my markdown using:

```
(perun/render :renderer 'com.andrearichiardi.blog.post/render
              :out-dir "blog/posts/")
```

with `:dir-root "public"`, I get permalinks like:

`/blog/posts/public/post1.html`
`/blog/posts/public/post2.html`

I am trying a couple of things, but [this](https://github.com/hashobject/perun/blob/master/src/io/perun.clj#L812) line seems suspicious. The order feels flipped, but maybe I am totally off and I will try a couple of things more first :smile: 

Any hint appreaciated!

arichiardi:
close, but not quite, there is more parsing to do in order to achieve the desired result. Hopefully patch coming soon.

PR #202, opened 2017-12-26
Add task for compiling less files

bhagany:
I was inspired to do this by a conversation with @arichiardi on slack - it's pretty basic, but I think it fits well with Perun as it stands.

### DIFF
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -1244,3 +1244,35 @@
         :passthru-fn content-passthru
         :task-name "inject-scripts"
         :tracer :io.perun/inject-scripts}))))
+
+(def ^:private +lessc-defaults+
+  {:out-dir "public"
+   :out-ext ".css"
+   :filterer identity
+   :extensions [".less"]
+   :include-dirs []})
+
+(deftask lessc
+  "Compile CSS using Less.
+
+  `lessc` must be installed and on your PATH."
+  [d out-dir      OUTDIR     str   "the output directory"
+   x out-ext      OUTEXT     str   "extension output"
+   _ filterer     FILTER     code  "predicate to use for selecting entries (default: `identity`)"
+   e extensions   EXTENSIONS [str] "extensions of files to process"
+   m meta         META       edn   "metadata to set on each entry"
+   i include-dirs PATH       [str] "Include directories for lessc."]
+  (let [options (merge +lessc-defaults+ *opts*)
+        include-dirs-str (->> (:include-dirs options)
+                              (mapcat (fn [dir]
+                                        (if (.startsWith dir "/")
+                                          [dir]
+                                          (map #(str % "/" dir) (boot/get-env :directories)))))
+                              (string/join ":"))]
+    (content-task
+     {:render-form-fn (fn [data] `(io.perun.lessc/compile-less ~data ~include-dirs-str))
+      :paths-fn #(content-paths % options)
+      :passthru-fn content-passthru
+      :task-name "lessc"
+      :tracer :io.perun/lessc
+      :rm-originals true})))
--- a/src/io/perun/lessc.clj
+++ b/src/io/perun/lessc.clj
@@ -1,0 +1,18 @@
+(ns io.perun.lessc
+  (:require [io.perun.core :as perun]
+            [boot.from.me.raynes.conch :as conch]))
+
+(defn compile-less
+  [{:keys [entry]} include-dirs]
+  (perun/report-debug "lessc" "compiling lessc" (:path entry))
+  (let [lessc-args (->> [(when (seq include-dirs)
+                           (str "--include-path=" include-dirs))
+                         (:full-path entry)]
+                        (keep identity))
+        p (apply conch/proc "lessc" lessc-args)
+        css (conch/stream-to-string p :out)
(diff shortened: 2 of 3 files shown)


podviaznikov (approved):

podviaznikov:
Code looks good to me

arichiardi:
Thanks for the nice gift @bhagany :christmas_tree: 

I am trying this, while it is working fine, I don't feel totally at easy with the following: in order to build now, I have basically to install `lessc` externally (not from `boot` I mean) - the build is not self-contained anymore.

So I still think that `boot-npm` provides value here, sorry if I am a bit picky :smile: just sharing some concern :wink:

arichiardi:
Actually there is a little problem, when I use it this way:

```
(perun/lessc :filterer #(-> % :filename (= "blog.less")))
```

I get:

```
$ tree target/public/

└── public
    └── css
        └── blog.css
```

Probably some issue with the output dirs, which I think I had to take into consideration in my task as well.

bhagany:
> I am trying this, while it is working fine, I don't feel totally at easy with the following: in order to build now, I have basically to install lessc externally (not from boot I mean) - the build is not self-contained anymore.

Yes, this is what I meant in our conversation on slack, here:
image

As for your output dir problem, I think Perun's pattern is just different from your expectation.  If your input file is at `public/css/blog.less`, and your `:out-dir` is `public` (which is the default), then what you're seeing is expected.  Perun takes the input path, applies any transformations like rendering, and then appends that to `:out-dir`.

I think if you want things in your input `public` dir to be in `target/public`, you can set `:out-dir` to `""`

arichiardi:
ok got it, so that probably is solved just by moving the folder out of `public`. It makes sense. Thanks I did not think about it ;)

bhagany:
... likewise, I'm rather inclined to say that this change isn't in line with how Perun works elsewhere https://github.com/hashobject/perun/pull/200/files

bhagany:
Ah, yes, moving your files out of `public` in the input would work as well

arichiardi:
Uhm, but that did not make too much sense. Why is `public` showing up after `out-dir`? The problem is better explained here https://github.com/hashobject/perun/issues/199

bhagany:
Yes, it’s another instance of the same misunderstanding, possibly complicated by `:dir-root`. To explain, this:
```
(perun/render :renderer 'com.andrearichiardi.blog.post/render
              :out-dir "blog/posts/")
```
should take an input file like `public/foo.html` and put the output at `blog/posts/public/foo.html` because of the appending behavior explained earlier. The permalink is just a function of that path, and `:dir-root` only comes into play if it exists at the beginning of the file’s path, in which case it gets removed. 

That change to `render` breaks the appending behavior that every other Perun task with file output still has.

arichiardi:
Uhm, we can of course revert the commit, if I can come up with an alternative solution :smile:

bhagany:
ah, yeah, without that commit the "Perun way" would be the same as with the less file - move the input file out of `public`

arichiardi:
ok, I am fine without the commit, I can always change my stuff ... betten than breaking everything for everybody :smile:

bhagany:
@podviaznikov any opinions on this?

podviaznikov:
First of all https://github.com/degree9/boot-npm looks impressive. Can't believe how many projects come up everytime in boot ecosystem. Does it require you to have node/npm installed?
If it requires node/npm installation - this seems to me similar to us requires less to be installed.

I guess I would go with this implementation.
The beauty of perun is that user is not forced to use this plugin but use your own version of it (with different implementation).

Core plugins seems like a bit opinionated batteries (we can change our opinions later too) - that everyone is free to replace/ignore.

Does it make sense to you @bhagany and @arichiardi? Just my opinion. Also, the cool thing about plugin - implementation can be changed in the future and task would work. So I'd rather merge this one

arichiardi:
I think you are right @podviaznikov, still something external needs to be installed.

bhagany:
Yes, that makes sense to me @podviaznikov

bhagany:
It just occurred to me that you could probably do something like this to install less for you:

```clj
(def my-lessc
    (comp (degree9.boot-npm/npm :install "less@latest" :global true)
          (io.perun/lessc)))
```

bhagany:
hmmmmmmmm hm hm hm - I am reconsidering my position on the `:out-dir` problem @arichiardi ... I found some old code that never got merged that addresses a similar issue, but for all of Perun instead of just `content-paths`.  I'll follow up in a separate PR.

arichiardi:
About lessc, I am using the [boot-css](https://github.com/degree9/boot-css/blob/master/src/degree9/boot_css.clj) task at the moment cause it basically does everything for you. It's been simplified considerably.

bhagany:
ah, yes, but Perun's will do fileset diffing, and skip less compilation if possible :)

arichiardi:
Oh, right, that's nice to have :smile: Big selling point!

arichiardi:
Hi folks, just wondering if this one is going to be merged. I have a new blog waiting for a blog post :smile:

bhagany:
nice! I like to give other people a bit of time to weigh in if they want, but if there's demand, I don't mind merging it :)

PR #203, opened 2017-12-30
Fix repetition of :out-dir when perun processes a file more than once

bhagany:
@arichiardi I think that this applies pretty much the same fix you merged in #200, but everywhere

### DIFF
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -157,6 +157,13 @@
           (commit tmp)
           (pm/set-meta updated-metas)))))
 
+(defn apply-out-dir
+  [path old-out-dir new-out-dir]
+  (let [path-args (if (= old-out-dir new-out-dir)
+                    [path]
+                    [new-out-dir path])]
+    (apply perun/create-filepath path-args)))
+
 (defn render-in-pod
   "Renders paths in `inputs`, using `render-form-fn` in `pod`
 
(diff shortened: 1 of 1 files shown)


arichiardi:
Ok cool so I will test this out against the current working site I have and report back

arichiardi:
So I checked this and now with the following:

```clojure
(def posts-dir "blog/posts")
(perun/render :out-dir posts-dir :renderer 'com.andrearichiardi.blog.post/render-page)
```

I get:

```shell
￼￼￼target/
├── blog
│   └── posts
│       └── public
│           ├── cljs-on-gh-pages.html
│           ├── cljs-replumb-require.html
│           ├── replumb-ast.html
│           └── semantic-perun-blog.html
└── public
    ├── blog
    ...
```

While before it was correctly under `public`, also note that all my posts are in `resources` (not `resources/public`). I feel this is expected so the patch seems to work for me.
Should I try something else just to make sure?

arichiardi:
Uhm, but now if I move the posts to `public` it is not rendering them anymore.

src/io/perun.clj
```diff
@@ -802,20 +808,17 @@
...
+  (let [{:keys [renderer out-dir] :as options} (merge +render-defaults+ *opts*)]
     (letfn [(render-paths [fileset]
               (let [entries (filter-meta-by-ext fileset options)]
                 (reduce
-                 (fn [result {:keys [out-dir filename path] :as entry}]
+                 (fn [result {:keys [path] :as entry}]
                    (let [content (slurp (boot/tmp-file (boot/tmp-get fileset path)))
-                         path-args (if (= out-dir (:out-dir options)) ;; if custom out-dir we cannot rely on pre-built :path
-                                     [path]
-                                     [out-dir (:out-dir options) filename])
```
arichiardi, line 813: this lines seems to me to build the path that would work for me, `public` before the specified `:out-dir`.

arichiardi:
So ok, I finally settled with another solution, blog posts are in `resources/blog/posts` and the render function is a simple:

```clojure
(perun/render :renderer 'com.andrearichiardi.blog.post/render-page)
```

podviaznikov (approved):

arichiardi:
It would be great to merge this one as well and maybe deploy a `SNAPSHOT` with both this and the `lessc` task :wink:

PR #206, opened 2018-01-08
Add an example of using perun with Semantic UI

arichiardi:
This also shows the problem in https://github.com/boot-clj/boot/pull/680, the `css` folder does not contain the `css/themes` folder when served using `boot-http` because it was moved with `sift`.


bhagany:
Is there any way we could reduce the number of files needed?

arichiardi:
Yes I will get rid of the highlight stuff. In general no, Semantic UI has that many.

arichiardi:
I tried once to use a git submodule but because of the bug you are trying to fix, `sift`-ing things around was not working.

bhagany:
dang, and there's no way to pare it down to only the ones this example needs?

arichiardi:
Yeah but that would need some more work. Probably just a bunch of files in `site` can be safely trimmed away

arichiardi:
I can always pull it out in a separate repo if you like

arichiardi:
@bhagany @podviaznikov would it be worth merging this one?

podviaznikov:
I guess we can merge it now and always extract examples into the separate repo later.
Seems like perun needs at least 3 repos now:
perun, perun-site, perun-examples.

podviaznikov (approved):

allentiak:
> I guess we can merge it now and always extract examples into the separate repo later.
> Seems like perun needs at least 3 repos now:
> perun, perun-site, perun-examples.

@podviaznikov 
Should I create a new repo to include the examples?

podviaznikov:
>@podviaznikov
> Should I create a new repo to include the examples?

@allentiak what do you think?

I think we are good for now since you removed `expose` example.
But I guess we can create new repo with bigger examples: like this one.

However, I'm also not sure if examples are that needed, since we have links to real world examples.

allentiak:
> @podviaznikov
> Should I create a new repo to include the examples?

@podviaznikov I wrote this before even starting #225.
As you say, given that #225 has been merged, there is no need for any separate repo for the examples anymore.

allentiak (commented):
Hi, @arichiardi ,
Thanks a lot for this PR!

@podviaznikov, @eccentric-j: What's your take on using JS code, as oppossed to using CLJS?

podviaznikov:
@allentiak yeah, I realized that timing was different:)

jaidetree (approved):
Looks alright to me. As far as JS vs CLJS, while I favor CLJS, perun's value is that it's the tool to dive in and build your own pipeline\static site generator vs something like hugo where you just change the template and rely on its existing functionality. So with that in mind, showing examples with JS may be valuable.

allentiak:
@arichiardi 

Have you tested this code?

arichiardi:
It taken from my working blog but I haven't tested it with new boot/perun

allentiak:
Ciao, Andrea,

In this case...
Could you...
1) rebase against current master?
2) test again your example?
3) If possible, could you please try to integrate your example into
the existing example? From a previous discussion on #225, we now have
one single, consolidated example. This makes it easier to maintain.

Thanks again!
Leandro

arichiardi:
@allentiak Ciao! Will do my best ;)

allentiak (changes requested):
(Wrapping up my previous comment, just for the record.)

Great idea!

Could you please:
1. rebase against current master?
2. test again your example?
3. try to integrate your example into the existing example?

Thanks again for your PR!

allentiak:
I will take care of finishing this one.

PR #208, opened 2018-01-21
Use lessc dependencies to trigger re-renders

bhagany:
It turns out that lessc has a relatively easy way of tracking dependencies for the files that it compiles, and we can leverage this to know when to recompile at a granular level.  There are no user-facing api changes.

### DIFF
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -1293,6 +1293,22 @@
    :extensions [".less"]
    :include-dirs []})
 
+(defn lessc-paths
+  [fileset options include-dirs-str pod]
+  (let [path-map (->> fileset
+                      pm/get-meta
+                      (map (juxt :full-path :path))
+                      (into {}))]
+    (reduce (fn [inputs [path {:keys [entry] :as input}]]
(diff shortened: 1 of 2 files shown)


arichiardi:
@bhagany first of all thanks as usual for working on this.

The patch achieves reloading of imported files, like `.variables` and `.overrides` for Semantic UI.
There is a problem though: at the end of the pipeline I see the following and the main css file does not seem to be updated:

```
Process completed successfully.
[lessc] - copied unchanged file public/css/blog.css
Sifting output files...
```

arichiardi:
This still has the problem described, seems like a simple fix, will do my best to check that

allentiak (changes requested):
(Just explicitly mentioning previous comments...)

> The patch achieves reloading of imported files, like `.variables` and `.overrides` for Semantic UI.
> There is a problem though: at the end of the pipeline I see the following and the main css file does not seem to be updated:
> 
> ```
> Process completed successfully.
> [lessc] - copied unchanged file public/css/blog.css
> Sifting output files...
> ```

arichiardi:
Will have to check this, after one year I forgot about it and of course it is a very nice to have 😀

PR #209, opened 2018-01-21
Replace list of plugins with reference to perun.io

jstaffans:
The list of built-in tasks in the README was not up-to-date. I'd suggest keeping only the list over at perun.io as a reference and remove it from the README.


bhagany (approved):
Good idea, thank you

PR #218, opened 2018-11-09
Sort files by publish date in RSS feed.

ballpointcarrot:
My rss feeds were generating entries in random order; this change allows
generating rss feeds consistently, sorted by the publish date set in the
entry.


podviaznikov (approved):
Makes sense, thank you!

PR #222, opened 2019-02-04
Added eccentric-j.com perun example to README.md

jaidetree:
Added my site as an example of a blog built with perun :metal:


podviaznikov:
thank you @eccentric-j!

PR #223, opened 2019-02-10
Fix support for Java 11

jaidetree:
## Updates
- Updated dependencies such as clj-yaml, pugets, and flexmark
- Updated markdown task as the flexmark options adapter changed
- Added a Dockerfile_java11 file to test the example blog build in Java 11
- Added a compose file to organize the docker builds.
- Updated changelog

## Docker Compose

1. Run `docker-compose run perun-java11`
2. It will attempt to build the image and then the blog example.

## If it fails
1. Make changes to the perun source 
2. Run `docker-compose run perun-java11 bash`
3. Run `cd /usr/src/perun`
4. Run `boot install`
5. Run `cd /usr/src/app`
6. Run `boot build` to rebuild the blog with updated perun lib

I'm hoping we can use the docker compose file to organize other builds for testing purposes as cases increase.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
   effectively the same. These options have been unified to `out-dir`. ([#83][83])
 
 [83]: https://github.com/hashobject/perun/issues/83
+[hashobject/perun.io#28](hashobject/perun.io#28): Update dependencies to support Java 11
 
 ## 0.3.0 (2016-01-17)
 
--- a/Dockerfile_java11
+++ b/Dockerfile_java11
@@ -1,0 +1,6 @@
+FROM clojure:openjdk-11-boot
+ADD . /usr/src/perun
+WORKDIR /usr/src/perun
+RUN boot build
+WORKDIR /usr/src/app
+CMD ["boot","build"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,9 @@
+version: '3'
+services:
+  perun-java11:
+    build:
+      context: .
+      dockerfile: Dockerfile_java11
+    volumes:
+      - ./examples/blog:/usr/src/app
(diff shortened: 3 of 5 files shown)


jaidetree:
Fixes hashobject/perun.io#28

allentiak:
@bhagany @podviaznikov
Would it be OK merging this?

@eccentric-j Thanks for the fix!
On the other hand, I'm not so sure whether it is convenient to publish a fork you don't seem to intend to maintain... If people start using it, you will have the pressure to keep it up to date...
Two published packages may mean two packages to keep up to date...
(Just a thought, though.)

jaidetree:
I needed something up earlier so I could get fixes in a circle-ci build. My plan is to kill my fork when these changes are merged but at this point it's got two compatibility fixes so it is maintained.

podviaznikov:
I think code looks good. I want to test it. Will do it today. Thank you for PR and discussion here.

jaidetree:
@podviaznikov Thanks! Please let me know if anything comes up in your tests.

podviaznikov:
@eccentric-j I think all looks good.
The only thing is that atom plugin didn't render anything for blog example: `perun-java11_1  | [atom-feed] - atom-feed found nothing to render`. But maybe it was broken before. Will need to investigate.
Otherwise lets merge!

Thank you for this!

jaidetree:
Great 👍  I looked into the atom feed, I think it's because examples/blog/build.boot is:

```clojure
        (perun/atom-feed :filterer :original)
```

https://github.com/hashobject/perun/blob/master/examples/blog/build.boot#L48

But the example posts don't have `:original true` in their yaml headers so it doesn't output anything.

podviaznikov:
@eccentric-j thank for investigating.
So it's something to look deeper.
`:original true` should be set automatically. So some other bug I guess

PR #225, opened 2019-02-11
Consolidate examples

allentiak:
WIP = Work In Progress :-)


jaidetree (commented):
Looks good to me, thanks for fixing the examples 👍 Let’s see what @podviaznikov has to say 😄

allentiak:
There are two examples: a blog and an exposé.
The first one works OK.

The second one is incomplete and buggy. As it stands now, it is only dead weight.

I think the best is to simply erase the exposé, leaving only the blog.
There are plenty of real-world, working websites in the main README.md.

What's your opinion, @podviaznikov , @eccentric-j ?

jaidetree (commented):
I agree with you, the expose is another example that will require maintenance. Plus if a person follows along the blog example I'm sure they'll get pretty quickly how to use a subset of those features for a simpler expose style site.

jaidetree (approved):

podviaznikov (approved):

podviaznikov:
Looks good! Thank you

eccentric-j:
Good work, thanks Allentiak.

jaidetree:
Good work, thanks @allentiak and @podviaznikov!

PR #236, opened 2019-05-23
Replaced the default slug-fn with a more robust version.

wiseman:
The previous default slug-fn used by the slug task assumed
Jekyll-style naming, e.g., "2001-01-01-my-post.md", and if this
assumption was violated it would return non-sensical results.

Previous slug-fn:

| Source filename         | Slug                | Final filename           |
|-------------------------|---------------------|--------------------------|
| "2001-01-01-my-post.md" | "my-post"           | "my-post.html"           |
| "index.md"              | ""                  | ".html"                  |
| "my-great-new-post.md"  | "post"              | "post.html"              |

The new default slug-fn handles Jekyll-compliant names in the same
way, but if given a filename that doesn't fit the Jekyll format it
will just strip the extension.

New slug-fn:

| Source filename         | Slug                | Final filename           |
|-------------------------|---------------------|--------------------------|
| "2001-01-01-my-post.md" | "my-post"           | "my-post.html"           |
| "index.md"              | "index"             | "index.html"             |
| "my-great-new-post.md"  | "my-great-new-post" | "my-great-new-post.html" |

I think think it's worth changing the default to a function that gives useful results in more cases, especially since stock perun doesn't use the date info in the filename for anything, making the Jekyll convention unnecessary.


allentiak:
Hi!
Thanks for your PRs!

I will have time to reviewing them in then upcoming days.

podviaznikov (approved):
@wiseman this looks great!
Thank you for PR. Especially for giving example of old and new behavior.
I'll approve it. Also if you have time, it will be cool if you can implement this desired behavior in unit tests. We do have some unit tests, so this should be doable - just test slug function itself.

allentiak:
@podviaznikov

Thanks for reviewing this!

arichiardi:
Should we bring this to completion? I can help with rebase very likely in the next couple of weeks

allentiak:
Not me, sorry.
https://github.com/hashobject/perun/issues/241#issuecomment-792000036

Issue #238, opened 2019-05-26
Gravatar link should be https


arichiardi:
Hello folks,

I am receiving:

```
aws-lambda-runtime-lumo.html:3 Mixed Content: The page at 'https://andrearichiardi.com/blog/posts/aws-lambda-runtime-lumo.html' was loaded over HTTPS, but requested an insecure image 'http://www.gravatar.com/avatar/708f9c347fce0493d929be88dbf70ea9'. This content should also be served over HTTPS.
```

From one of my blog posts and I was wondering if there is a way to set `https` to the gravatar link.

I have seen tests have `http` but could not find where to change it? Is it baked in their library?

I can provide a patch with a bit of guidance.

Thanks!

podviaznikov:
Hi @arichiardi. Yes, you can provide patch.

Seems like it's here https://github.com/hashobject/perun/blob/master/src/io/perun/gravatar.clj#L7 and the library we use provides `https` support https://github.com/alesguzik/gravatar-clj/blob/master/src/gravatar/core.cljc#L16.

So you can switch it to `https` for everyone.

arichiardi:
Cool thank you for your pointers! I will definitely do it.

PR #242, opened 2020-01-14
Default to https for gravatar links

arichiardi:
It was old-style http and we want to default to https instead.

This also introduces the new `boot test` task with the right setup for
reporting and exiting correctly in case of failure.


arichiardi:
Finally I got around to do this :smile:

allentiak:
Sorry for the late reply, @arichiardi.
Thanks for your hard work!
I'll check it.

allentiak:
Thanks! Sorry for the delay.

allentiak:
Closes #238

PR #250, opened 2020-10-21
Feat/rss with author

munen:


munen:
Args.. I apologize for the spam. GH always makes a PR against upstream instead of our fork.

At some point, it might make sense to open a PR against upstream. But for the moment, we're figuring out ourselves on how to manage feeds properly.

PR #251, opened 2020-10-21
Fix: RSS and Atom Feeds

munen:
Both Atom and RSS feeds had some issues. This PR is not a perfect solution, but it improves the situation quite a bit.

Changes for Atom:

- Atom feeds were not valid before. The rather old version of `clojure.data.xml` library would render `<feed xmlns:a="http://www.w3.org/2005/Atom">` instead of `<feed xmlns="http://www.w3.org/2005/Atom">` (notice the `:a` namespace). Whilst this doesn't look like a huge issue, it leads to various issues:

  - https://validator.w3.org/feed/ would not deem the feed valid

image

  - The feed would be rejected by various feed readers (i.e. Thunderbird and Miniflux).

image

  - Other feed readers would show 'random' errors. For example feedbin.com will skip various posts whilst showing others.

The issue was not trivial to fix, because the docs of `clojure.data.xml` state namespaces as a new feature. And this new syntax would not work on the old version. Finally, I realized that upgrading to the current version would give me access to said syntax. And since atom.clj is not specifying any namespace, now none is rendered and will make the Atom feed valid.

Unfortunately, the tests are now breaking. They run very slow for me and I don't know how to debug them. Hence, I left them broken (apologies). Simply reverting the clojure.data.xml version will make the tests run again, but then the feed will not be valid.

If someone teaches me how to run perun with Cider so that I can do some debugging or run just one test, I'd be happy to give that a shot at some point.

Changes for RSS:

Firstly, they are not nearly as huge as the Atom issue(;

- The sort order of posts in many feed readers was random, because the sort order happend on the name of the week day instead of the date. There's more information inline in the code.

- It would be nice to have included the complete content of the post into the RSS feed, but I'm lacking the necessary knowledge to do that in Perun. The `:content` attribute for a post is not accessible in rss.clj; it is in atom.clj, though. I've added the necessary boilerplate.

- It would be nice to support multiple authors, but since the used clj-rss library does not support it, I've only left informative comments for the next dev checking that out.

- The RSS spec allows for "author" and "author-email" to be given as meta-date. I've added the specified syntax.

---

As a result, both RSS and Atom feeds for 200ok.ch are valid now:

- https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2F200ok.ch%2Fatom.xml
- https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2F200ok.ch%2Frss.xml

Both still have recommondations that could be addressed, but at least they are valid and work for all tested feed readers - including the ones that silently didn't work for one reason or another(;

### DIFF
--- a/build.boot
+++ b/build.boot
@@ -2,6 +2,7 @@
   :source-paths #{"test"}
   :resource-paths #{"src"}
   :dependencies '[[boot/core "2.8.2" :scope "provided"]
+                  [org.clojure/clojure "1.10.1"]
                   [adzerk/boot-test "1.2.0" :scope "test"]
                   [adzerk/bootlaces "0.2.0" :scope "test"]])
 
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -1182,7 +1182,7 @@
 
 (def ^:private ^:deps atom-deps
   '[[org.clojure/tools.namespace "0.3.1"]
-    [org.clojure/data.xml "0.0.8"]
+    [org.clojure/data.xml "0.2.0-alpha6"]
     [clj-time "0.15.2"]])
 
 (def ^:private +atom-defaults+
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -1228,7 +1228,7 @@
       :pod (create-pod atom-deps)})))
 
 (def ^:private ^:deps rss-deps
-  '[[clj-rss "0.2.5"]
+  '[[clj-rss"0.2.6"]
     [clj-time "0.15.2"]])
 
 (def ^:private +rss-defaults+
--- a/src/io/perun.clj
+++ b/src/io/perun.clj
@@ -1182,7 +1182,7 @@
 
 (def ^:private ^:deps atom-deps
   '[[org.clojure/tools.namespace "0.3.1"]
-    [org.clojure/data.xml "0.2.0-alpha6"]
+    [org.clojure/data.xml "0.0.8"]
     [clj-time "0.15.2"]])
 
 (def ^:private +atom-defaults+
--- a/src/io/perun/atom.clj
+++ b/src/io/perun/atom.clj
@@ -42,48 +42,48 @@
         atom (xml/emit-str
               (xml/sexp-as-element
                [::atom/feed {:xmlns "http://www.w3.org/2005/Atom"}
-                [:title site-title]
+                [::atom/title site-title]
                 (when (seq description)
-                  [:subtitle description])
-                [:generator {:uri "https://perun.io/" :version version} "Perun"]
-                [:link {:href base-url :type "text/html"}]
-                [:link {:href canonical-url :rel "self"}]
-                [:link {:href (:first navs) :rel "first"}]
-                [:link {:href (:last navs) :rel "last"}]
+                  [::atom/subtitle description])
+                [::atom/generator {:uri "https://perun.io/" :version version} "Perun"]
+                [::atom/link {:href base-url :type "text/html"}]
+                [::atom/link {:href canonical-url :rel "self"}]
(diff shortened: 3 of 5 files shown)


allentiak:
Thanks for the feature, @munen !

That being said, I think merging it with a broken test would be a pity...
If you want, we could meet to do some pair programming in the upcoming weeks, so we give it a look? Maybe if we work together we can make it work...
I haven't worked on this code for a long time...

allentiak (changes requested):
(See [conversation message](https://github.com/hashobject/perun/pull/251#issuecomment-713909511))

munen:
Thank you for inviting me to this repository @allentiak :pray: 

> That being said, I think merging it with a broken test would be a pity...

I agree. It's not even just a broken test, that I probably could have fixed. It's crashing with a stack trace. Hence, debugging would be benefitial.

For our site (200ok.ch), I've received mails from readers that they couldn't read the blog via Atom (which we link to in the header). My main priority was to get that to work and then to see if there's interest upstream - which I am very happy to see(;

> If you want, we could meet to do some pair programming in the upcoming weeks, so we give it a look? Maybe if we work together we can make it work...

That's a very kind offer which I'd happily take you up on :pray: 

Are you currently in UTC-3? If so, then we are five hours apart (I'm at UTC+2). I'm currently doing mostly product work, so I don't have too many strict appointments. For example, I could do today or tomorrow in our afternoon, so your morning.

> I haven't worked on this code for a long time...

No worries. I know that you've reached out for help on this project. I haven't stepped up for two reasons mainly: There's lots of other stuff (I'm not the only one here, though), and I'm not what I personally would call familiar with boot or the Perun code base. Having said that, we've invested quite a bit into building 200ok.ch based on Perun. So I do have a high motivation to fix bugs and keep the project going, because otherwise at some point we would have to migrate away. And rebuilding doesn't always/often make things better(;

allentiak:
@munen Thanks for your message! I've replied you privately.

munen:
Thanks to an answer of [Alex Miller himself on SO](https://stackoverflow.com/questions/64611482/how-to-generate-a-xml-without-a-namespace-with-the-latest-version-of-org-clojure/64615916#64615916), the tests now run and we have a feed without an auto-generated prefix.

However, the feed arguably has become become valid, but moved further into the invalid:

image

munen:
I have updated the question on SO and answered Mr. Alex Miller: https://stackoverflow.com/questions/64611482/how-to-generate-a-xml-without-a-namespace-with-the-latest-version-of-org-clojure

allentiak:
I saw you got some interesting replies to your question ;-)

munen:
> I saw you got some interesting replies to your question ;-)

Yes, so :crossed_fingers: that it'll get resolved there, too(;

allentiak:
> we have a feed without an auto-generated prefix.

@munen That certainly *is* progress :-)


> However, the feed [...] moved further into the invalid

@munen I saw the error messages, but... What do you mean by this, exactly? Is this the validator bug you mentioned in your SO question?

munen:
> @munen I saw the error messages, but... What do you mean by this, exactly? Is this the validator bug you mentioned in your SO question?

I wouldn't call it a 'validator bug' which implies in a bug in the validator, but yes I mean the update that I've added to the SO question where I go into further detail. With the proposed fix from Alex Miller, the feed actually has more errors, even though the first error went away.

allentiak:
I see. Thanks for the explanation, @munen!

munen:
Tests are green now:

> Ran 6 tests, 94 assertions, 0 failures, 0 errors.

Also, both RSS and Atom feeds are valid.

This PR is ready to merge.

allentiak (approved):

allentiak:
@munen  Thank you very much for your hard work!

Merged :-)

